PR #13, opened 2018-01-11
Charles/rapsearch index

cdebourcy:


boris-dimitrov:
Does the output path contain version information?    Is it possible with this approach for the upload to S3 to be overwriting an existing index that is at that moment being pulled from S3 by an machine init script?

cdebourcy:
No, the output path does not contain version information. This is work in progress; right now we're just overwriting the existing index while updating the version tracker file. I'm planning to add archiving of existing indexes in the future but that will be a separate PR. Thanks for raising my attention to these issues!

PR #16, opened 2018-01-13
Charles/version output

cdebourcy:
Produce an output file containing versioning information for a pipeline run.

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -13,6 +13,7 @@ NCBITOOL_S3_PATH = "s3://czbiohub-infectious-disease/ncbitool" # S3 location of 
 
 STATS = []
 LOGGER = None
+OUTPUT_VERSIONS = []
 
 class Updater(object):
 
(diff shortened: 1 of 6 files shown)


idseq_pipeline/commands/common.py
```diff
@@ -171,29 +172,49 @@ def get_remaining_reads_from_stats():
...
     LOGGER.info("========== %s ==========" % logparams.get("title"))
+
     # copy log file -- start
     LOGGER.handlers[0].flush()
     execute_command("aws s3 cp %s %s/;" % (LOGGER.handlers[0].baseFilename, logparams["sample_s3_output_path"]))
+
+    # record version of any reference index used
+    version_file_s3 = logparams.get("version_file_s3")
+    if version_file_s3:
+        version_json = execute_command_with_output("aws s3 cp %s -" % version_file_s3)
```
boris-dimitrov: nice

idseq_pipeline/commands/common.py
```diff
@@ -171,29 +172,49 @@ def get_remaining_reads_from_stats():
...
+        OUTPUT_VERSIONS.append(version_json)
+
+    # upload version output
+    output_version_file = logparams.get("output_version_file")
+    if output_version_file:
+        with open(output_version_file, 'wb') as f:
+            json.dump(OUTPUT_VERSIONS, f)
+        execute_command("aws s3 cp %s %s/;" % (output_version_file, logparams["sample_s3_output_path"]))
```
boris-dimitrov: if you wanted to stream into aws s3 cp, you could do it with python subprocess here... using the PIPE args

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -68,6 +69,13 @@
 DEFAULT_LOGPARAMS = {"sample_s3_output_path": SAMPLE_S3_OUTPUT_PATH,
                      "stats_file": os.path.join(RESULT_DIR, STATS_OUT)}
 
+# versioning
+## For now, index updates are infrequent and we can get their versions from S3.
+## If updates ever become frequent, we may want to check instead which version is actually on the
+## machine taking the job, possibly out of sync with the newest version in S3. 
+GSNAP_VERSION_FILE_S3 = 's3://czbiohub-infectious-disease/references/nt_k16.version.txt'
```
boris-dimitrov: I may have mentioned something to this effect before, but this isn't safe when multiple versions could be in use at the same time (even when transitioning between versions).

idseq_pipeline/commands/common.py
cdebourcy: Ah! That's cool!

PR #17, opened 2018-01-16
small change to chunk part numbering

boris-dimitrov:
since the number of parts can be determined in advance, we can use a slightly
more conventional naming scheme and still have lexicographic order

also adding some rudimentary validation

### DIFF
--- a/idseq_pipeline/commands/non_host_alignment_functions.py
+++ b/idseq_pipeline/commands/non_host_alignment_functions.py
@@ -48,7 +48,7 @@ UNIDENTIFIED_FASTA_OUT = 'unidentified.fasta'
 COMBINED_JSON_OUT = 'idseq_web_sample.json'
 LOGS_OUT_BASENAME = 'log'
 STATS_OUT = 'stats.json'
-VERSION_OUT = 'versions.json' 
+VERSION_OUT = 'versions.json'
 
 # arguments from environment variables
 FASTQ_BUCKET = os.environ.get('FASTQ_BUCKET')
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -501,15 +501,26 @@ def check_s3_file_presence(s3_path):
       return 0
 
 # job functions
-def chunk_input(input_files_basenames, chunk_nlines, part_suffix):
+def chunk_input(input_files_basenames, chunk_nlines, chunksize):
```
boris-dimitrov: Btw I tested this function in a standalone python interpreter.   Caught one bug, fixed.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -605,8 +616,7 @@ def run_gsnapl_remotely(input_files, lazy_run):
     remote_index_dir = "%s/share" % remote_home_dir
     # split file:
     chunk_nlines = 2*GSNAPL_CHUNK_SIZE
-    part_suffix = "-chunksize-%d-part-" % GSNAPL_CHUNK_SIZE
```
cdebourcy: This broke it, part_suffix is also used in run_gsnapl_remotely

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -708,8 +718,7 @@ def run_rapsearch2_remotely(input_fasta, lazy_run):
     remote_index_dir = "%s/references/nr_rapsearch" % remote_home_dir
     # split file:
     chunk_nlines = 2*RAPSEARCH_CHUNK_SIZE
-    part_suffix = "-chunksize-%d-part-" % RAPSEARCH_CHUNK_SIZE
```
cdebourcy: part_suffix is also used in run_rapsearch2_remotely

boris-dimitrov: oh sorry

cdebourcy: No worries :)

PR #19, opened 2018-01-18
Charles/subsampling

cdebourcy:
If SUBSAMPLE is set, subsamples inputs to specified number of read pairs after host filtering and before gsnap/rapsearch.

### DIFF
--- a/idseq_pipeline/commands/non_host_alignment_functions.py
+++ b/idseq_pipeline/commands/non_host_alignment_functions.py
@@ -52,7 +52,7 @@ STATS_OUT = 'stats.json'
 VERSION_OUT = 'versions.json' 
 
 # arguments from environment variables
-SUBSAMPLE = int(os.environ.get('SUBSAMPLE')) # number of read pairs to subsample to, before gsnap/rapsearch
+SUBSAMPLE = int(os.environ.get('SUBSAMPLE')) if os.environ.get('SUBSAMPLE') else None # number of read pairs to subsample to, before gsnap/rapsearch
 FASTQ_BUCKET = os.environ.get('FASTQ_BUCKET')
 INPUT_BUCKET = os.environ.get('INPUT_BUCKET')
 FILE_TYPE = os.environ.get('FILE_TYPE', 'fastq.gz')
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -108,6 +113,78 @@
...
+       sequence_basename = sequence_name.rstrip().rsplit('/', 1)[0]
+       if sequence_basename in read_ids_to_keep:
+            output.write(sequence_name)
+            output.write(sequence_data)
+            kept_read_ids += sequence_basename
+       sequence_name = input_fasta_f.readline()
+       sequence_data = input_fasta_f.readline()
+    input.close()
+    output.close()
+    assert set(kept_read_ids) == set(read_ids_to_keep), "Not all desired read IDs were found in the file: {}".format(input_file)
```
boris-dimitrov: This function is almost identical to the previous one.   Could be nice to unify them.   Also, in python it's better to use "with" statement with files, handles resource management cleanly regardless of exceptions etc.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -108,6 +113,78 @@
...
+
+def subsample_fastas(input_files, merged_file, target_n_reads):
+    total_records = 0.5 * count_lines_in_paired_files(input_files) # each fasta record spans 2 lines
+    # note: target_n_reads and total_records really refer to numbers of read PAIRS
+    if total_records <= target_n_reads:
+        return
+    subsample_prefix = "subsample_%d" % target_n_reads
+    records_to_keep = set(random.sample(xrange(total_records + 1), target_n_reads))
```
boris-dimitrov: Using `set` here negates the efficiency of random.sample acting as a generator, by creating a big data structure in memory.   If we are concerned these files could be really large --- millions of lines -- this could be a concern, but not greater than the concern of actually reading the files line by line, which is still likely to dominate.   In other words, doing this in python is likely not great for subsampling truly large files.

boris-dimitrov: https://en.wikipedia.org/wiki/Reservoir_sampling

PR #20, opened 2018-01-18
minor improvement, remove sleep after each gsnapl chunk

boris-dimitrov:
when rerunning sample 1039, a total of 4 hours and 25 minutes was
spent in these sleep(10) statements, which was unnecessary


boris-dimitrov:
We can also do this for rapsearch after this other pull request is merged:  https://github.com/chanzuckerberg/idseq-pipeline/pull/18

PR #23, opened 2018-01-24
small improvements to permit concurrency of chunks

boris-dimitrov:
so far I only did it for rapsearch so you can also compare to gsnap to see how it was before


idseq_pipeline/cli.py
```diff
@@ -48,19 +48,19 @@
...
     """Main CLI entrypoint."""
-    import idseq_pipeline.commands
+    import commands as idseq_pipeline_commands
     options = docopt(__doc__, version=VERSION)
 
     # Match user-entered command
     for (k, v) in options.items(): 
-        if hasattr(idseq_pipeline.commands, k) and v:
-            module = getattr(idseq_pipeline.commands, k)
-            idseq_pipeline.commands = getmembers(module, isclass)
```
boris-dimitrov: This line actually appears to have been modifying the loaded module, which is probably not intentional and slightly dicey.   I hope it was not intentional.   With the change everything seems to work fine.

cdebourcy: This is actually something I copied from skele-cli. Probably good to check if installation and CLI usage still works

boris-dimitrov: Heh ok!

boris-dimitrov:
Hey @cdebourcy do these results look OK?   I think something small is missing, but mostly there, first run through the parallel steps of stage 2:   ```aws s3 ls s3://czbiohub-idseq-samples-development/samples/314314314/7777777/results/```

cdebourcy:
@boris-dimitrov Doesn't look like there are results for any of the steps after rapsearch?
e.g. 
'filter.deuterostomes.taxids.rapsearch2.filter.deuterostomes.taxids.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.m8'
'unidentified.fasta'
'idseq_web_sample.json' (the main output that goes into the idseq-web report)

boris-dimitrov:
Stage 2 now works, both from scratch and lazily.

boris-dimitrov:
I believe this works now;  but my only end-to-end attempt failed due to an operational issue on one of the rapsearch instances (the pipeline code here operated as designed).

For some reason resque isn't picking up the job on rerun:   https://alpha.idseq.net/samples/282

boris-dimitrov:
...and here is the first end-to-end success:  https://alpha.idseq.net/samples/282

boris-dimitrov:
I also just submitted the infamous sample 1039 from prod:  https://alpha.idseq.net/samples/283

It's a big one, and there are already 7 jobs running on the system, so it will be interesting to see how it performs on this.

PR #41, opened 2018-02-01
misc cleanup

boris-dimitrov:
quiet all aws s3 cp commands;  we already print the command itself in the log

### DIFF
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -239,7 +239,7 @@ def run_star(fastq_files):
     # check if genome downloaded already
     genome_file = os.path.basename(STAR_GENOME)
     if not os.path.isfile("%s/%s" % (REF_DIR, genome_file)):
-        execute_command("aws s3 cp %s %s/" % (STAR_GENOME, REF_DIR))
+        execute_command("aws s3 cp --quiet %s %s/" % (STAR_GENOME, REF_DIR))
         execute_command("cd %s; tar xvfz %s" % (REF_DIR, genome_file))
         write_to_log("downloaded index")
     # Check if parts.txt file exists, if so use the new version of (partitioned indices). Otherwise, stay put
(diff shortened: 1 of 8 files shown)


idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -236,39 +236,37 @@ def run_star_part(output_dir, genome_dir, fastq_files):
...
+    star_outputs = [STAR_OUT1, STAR_OUT2]
+    num_fastqs = len(fastq_files)
+    def unmapped_files_in(some_dir):
+        return ["%s/Unmapped.out.mate%d" % (some_dir, i+1) for i in range(num_fastqs)]
+    genome_name = os.path.basename(STAR_GENOME).rstrip(".gz").rstrip(".tar")
+    if genome_name != "STAR_genome":
+        with print_lock:
+            print "Oh hello interesting new genome {}".format(genome_name)
+    genome_dir = os.path.join(REF_DIR, genome_name)
+    if not os.path.exists(genome_dir):
```
yunfangjuan: this could potentially be an issue if different hosts are sharing the same genome dir

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -236,39 +236,37 @@ def run_star_part(output_dir, genome_dir, fastq_files):
...
 def run_star(fastq_files):
-    # check if genome downloaded already
-    genome_file = os.path.basename(STAR_GENOME)
-    if not os.path.isfile("%s/%s" % (REF_DIR, genome_file)):
-        execute_command("aws s3 cp %s %s/" % (STAR_GENOME, REF_DIR))
-        execute_command("cd %s; tar xvfz %s" % (REF_DIR, genome_file))
+    star_outputs = [STAR_OUT1, STAR_OUT2]
+    num_fastqs = len(fastq_files)
+    def unmapped_files_in(some_dir):
+        return ["%s/Unmapped.out.mate%d" % (some_dir, i+1) for i in range(num_fastqs)]
```
yunfangjuan: Please verify that when the input is only 1 fastq file, the output is still Unmapped.out.mate{1/2} 
It's possible it's just Unmapped.out

boris-dimitrov: Interesting.   I just preserved the original semantics here, that was what the code before my change was assuming.

idseq_pipeline/commands/host_filtering_functions.py
boris-dimitrov: yes, it could!   i am happy to address that issue and also switch us into an environment where that sharing can happen;   and by different hosts i think realistically what we mean here is different jobs on the same instance --- because if you mean using a shared filesystem on amazon, I benchmarked EFS and it's worthless

boris-dimitrov: this said, it should be handle separately throughout the pipeline

i am going to refactor quite a bit to make it possible to address those issues

i only looked into this first because of the terrible performance of downloading 25GB index;   then found the correctness bugs that had to be fixed

boris-dimitrov:
Since this does contain an important bugfix I'd like to commit it as soon as it completes an end to end run.   I have submitted job:   https://alpha.idseq.net/samples/289

idseq_pipeline/commands/host_filtering_functions.py
cdebourcy: I checked it and it is unmapped.out.mate1 for single-end case

boris-dimitrov:
The test job already completed STAR successfully:   https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/91169d8f-8729-49a5-adec-a78d3c9851c5

boris-dimitrov:
How do I get a job to do multipart?   The one above used single part index.

PR #46, opened 2018-02-02
fix performance bug

boris-dimitrov:
the mutexes were always meant to be per service (i.e. one for gsnap
and another one for rapsearch) and i noticed that jobs from the
different services were waiting after each other because accidentally
the mutexes were per environment (which is the same for all services)


boris-dimitrov:
Test sample https://alpha.idseq.net/samples/303

boris-dimitrov:
Test sample worked.

PR #50, opened 2018-02-06
Curate accession2taxid file

cdebourcy:


idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,102 @@
...
+    print "Read the nt/nr file"
+    # Read the nt/nr file
+    dbids = {}
+    lines = 0
+    for gzf in [nr_file, nt_file]:  
+        with gzip.open(gzf, 'rb') as seqf:
+            for line in seqf:
+                lines += 1
+                if lines % 100000 == 0:
+                    print "%f M lines. " %  (lines/1000000.0)
```
boris-dimitrov: You might try `%3.1f` instead of `%f`.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,102 @@
...
+                    if s:
+                        dbids[s.group(1)] = 1
+    print "Read the mapping file and select ..."
+    # Read the accession2taxid mapping files
+    outf = open(output_mapping_file, 'wb')
+    lines = 0
+    for mapping_file in mapping_files:
+        with gzip.open(mapping_file, 'rb') as mapf:
+            for line in mapf:
+                fields = line.split("\t")
```
boris-dimitrov: If you only use the first 3 fields, performance may be marginally improved by providing a second argument to `split`, like `line.split("\t", 3)`.

boris-dimitrov: You are using a dictionary as a set.   I imagine the memory use might be slightly better if you actually use a set.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,102 @@
...
+        arguments = self.options
+        nt_file_local, nt_version_number = download_reference_locally_with_version_any_source_type(arguments['--nt_file'], dest_dir, dest_dir)
+        nr_file_local, nr_version_number = download_reference_locally_with_version_any_source_type(arguments['--nr_file'], dest_dir, dest_dir)
+        mapping_files_local = []
+        mapping_version_numbers = []
+        for f in arguments['--mapping_files'].split(","):
+            mapping_file_local, mapping_version_number = download_reference_locally_with_version_any_source_type(f, dest_dir, dest_dir)
+            mapping_files_local.append(mapping_file_local)
+            mapping_version_numbers.append(mapping_version_number) 
+        print "Reference download finished"
```
boris-dimitrov: If the reference downloads take time, they would probably take less time if they run concurrently in threads.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,102 @@
...
+        # Produce the output file
+        print "Curating accession2taxid"
+        output_mapping_file = os.path.join(DEST_DIR, 'curated_accession2taxid.txt')
+        curate_taxon_dict(nt_file_local, nr_file_local, mapping_files_local, output_mapping_file)
+        print "Curation finished"
+
+        # Convert to a berkeley db
+        print "Writing berkeley db"
+        output_db_file = os.path.join(DEST_DIR, 'curated_accession2taxid.db')
+        generate_accession2taxid_db(output_mapping_file, output_db_file, False)
```
boris-dimitrov: Do we need to erase the output file first?   In case it's a corrupt file from a failed prior run?   Not sure what the retry strategy will be.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,102 @@
...
+        output_s3_path = arguments['--output_s3_folder'].rstrip('/')
+        execute_command("gzip %s; aws s3 cp --quiet %s.gz %s/" % (output_db_file, output_db_file, output_s3_path))
```
boris-dimitrov: Overlap copy with compression:  ```"gzip -c {output_db_file} | aws s3 cp --quiet - {output_s3_path}/{output_db_file}.gz".format(output_db_file=output_db_file, output_s3_path=output_s3_path)```

Also if we switch to python3 we can use interpolation like #{} in Ruby so as not to repeat output_s3_path.

boris-dimitrov:
Nice work so far.

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: ok

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: ok

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: Yes - thanks!!

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: thanks

yunfangjuan:
Do you plan to compute the difference of different versions somewhere?

cdebourcy:
@yunfangjuan Yes -- adding it to this same command now

cdebourcy:
@yunfangjuan Accession ID diff is now uploaded to s3 too

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,159 @@
...
+        threads['nt'].start()
+        threads['nr'] = MyThread(download_reference_locally_with_version_any_source_type,
+            (arguments['--nr_file'], dest_dir, dest_dir))
+        threads['nr'].start()
+        mapping_files_local = []
+        mapping_version_numbers = []
+        mapping_files_sources = arguments['--mapping_files'].split(",")
+        mapping_file_results = {}
+        for f in mapping_files_sources:
+            threads[f] = MyThread(target=download_reference_locally_with_version_any_source_type,
```
boris-dimitrov: A bit paranoid here but won't hurt to `assert f not in threads`.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,159 @@
...
+import shelve
+import argparse
+import gzip
+import re
+import threading
+import traceback
+
+class MyThread(threading.Thread):
+    def __init__(self, target, args):
+        super(MyThread, self).__init__()
```
boris-dimitrov: ah you found that :)

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,159 @@
...
+        mapping_files_local = []
+        mapping_version_numbers = []
+        mapping_files_sources = arguments['--mapping_files'].split(",")
+        mapping_file_results = {}
+        for f in mapping_files_sources:
+            threads[f] = MyThread(target=download_reference_locally_with_version_any_source_type,
+                args=(f, dest_dir, dest_dir))
+            threads[f].start()
+        for f in threads:
+            threads[f].join()
```
boris-dimitrov: I'd add here `assert not threads[f].exception`.

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,159 @@
...
+import random
+import pickle
+import shelve
+import argparse
+import gzip
+import re
+import threading
+import traceback
+
+class MyThread(threading.Thread):
```
boris-dimitrov: I imagine given our current architecture this belongs in common.py

boris-dimitrov:
Nice to see this getting better.

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: ok

idseq_pipeline/commands/curate_accession2taxid.py
```diff
@@ -0,0 +1,143 @@
...
+        # Record difference with old accession2taxid
+        previous_mapping_s3 = arguments.get('--previous_mapping')
+        if previous_mapping_s3:
+            print "Computing diff"
+            previous_mapping_local = os.path.join(dest_dir, "previous_mapping.db.gz")
+            execute_command("aws s3 cp --quiet %s %s" % (previous_mapping_s3, previous_mapping_local))
+            execute_command("gunzip %s" % previous_mapping_local)
+            previous_mapping = shelve.open(os.path.splitext(previous_mapping_local)[0])
+            new_mapping = shelve.open(output_db_file)
+            previous_accessionids = set(previous_mapping.keys())
```
yunfangjuan: don't you need gigantic memory for this if each set? You can also loop through the new ones and see what's missing in the old one.

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: ok

idseq_pipeline/commands/curate_accession2taxid.py
cdebourcy: Ok, I'll submit another diff to convert to loop. To be clear: opening a python shelf and writing to it doesn't keep the entire map in memory right? It writes to the file while keeping very little in memory? Because otherwise we would need very large memory in the other steps anyway to produce the shelves.

PR #51, opened 2018-02-08
reduce pylint warnings

boris-dimitrov:
also change cp to ln in run_priceseqfilter


boris-dimitrov:
End to end test in https://alpha.idseq.net/samples/335

PR #52, opened 2018-02-09
skip deutero filter

cdebourcy:
goes with https://github.com/chanzuckerberg/idseq-web/pull/639

### DIFF
--- a/idseq_pipeline/commands/non_host_alignment_functions.py
+++ b/idseq_pipeline/commands/non_host_alignment_functions.py
@@ -65,7 +65,7 @@ STATS_OUT = 'stats.json'
 VERSION_OUT = 'versions.json'
 
 # arguments from environment variables
-SKIP_DEUTERO_FILTER = os.environ.get('SKIP_DEUTERO_FILTER')
+SKIP_DEUTERO_FILTER = int(os.environ.get('SKIP_DEUTERO_FILTER', 0))
 SUBSAMPLE = os.environ.get('SUBSAMPLE') # number of read pairs to subsample to, before gsnap/rapsearch
 FASTQ_BUCKET = os.environ.get('FASTQ_BUCKET')
 INPUT_BUCKET = os.environ.get('INPUT_BUCKET')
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -65,6 +65,7 @@
 VERSION_OUT = 'versions.json'
 
 # arguments from environment variables
+SKIP_DEUTERO_FILTER = os.environ.get('SKIP_DEUTERO_FILTER')
```
boris-dimitrov: I'd recommend
```
SKIP_DEUTERO_FILTER = int(os.environ.get('SKIP_DEUTERO_FILTER', 0))
```

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -1097,24 +1098,28 @@ def run_gsnap_and_dependent_steps():
...
-            DEFAULT_LOGPARAMS,
-            {"title": "filter deuterostomes from m8__1", "count_reads": True,
-             "before_file_name": os.path.join(RESULT_DIR, ANNOTATE_GSNAPL_M8_WITH_TAXIDS_OUT), "before_file_type": "m8",
-             "after_file_name": os.path.join(RESULT_DIR, FILTER_DEUTEROSTOMES_FROM_NT_M8_OUT), "after_file_type": "m8"})
-        run_and_log_eager(
-            logparams, TARGET_OUTPUTS["run_filter_deuterostomes_from_m8__1"],
-            run_filter_deuterostomes_from_m8,
-            os.path.join(RESULT_DIR, ANNOTATE_GSNAPL_M8_WITH_TAXIDS_OUT),
-            os.path.join(RESULT_DIR, FILTER_DEUTEROSTOMES_FROM_NT_M8_OUT))
+        if SKIP_DEUTERO_FILTER and int(SKIP_DEUTERO_FILTER) == 1:
```
boris-dimitrov: ...then here can be just
```
if SKIP_DEUTERO_FILTER:
   ...
```

yunfangjuan:
This will only be run for certain specific samples?

yunfangjuan:
Perhaps we should just skip it for all the samples and filter on the web app to be more flexible?

cdebourcy:
I was thinking about making it a filter in web app too (like blacklist), but didn't we decide the list was too long for that? It is still a great idea.
I think it makes a lot of sense to have it be a property of the host_genome though: conceptually, it's definitely a part of host_filtering

yunfangjuan:
Actually it's not so hard. We can filter it directly from the taxon_lineage table, right? Isn't deuterstroms at like phylum level or something?

cdebourcy:
Conceptually, yes. Except it's not one of the levels we're currently considering: it's superphylum.

PR #53, opened 2018-02-09
Fix bug that counted the total 2x even if input was single-end

cdebourcy:


boris-dimitrov:
This looks good.    Just noticed something else in that area that could be improved a bit.

```
    ...
    elif file_type == "m8" and line[0] == '#':
        continue
    else:
       count += 1
```
should probably be
```
    elif file_type != "m8" or line.starswith('#'):
       count += 1
```
Note particularly the replacement of `line[0] == '#'` with `line.startswith('#')`.   The former will break for empty lines.

PR #57, opened 2018-02-16
small perf improvements and bug fixes

boris-dimitrov:
small perf improvements and bug fixes
    
    * shorter connection timeout so dead machines don't hold up progress
    
    * double dispatch rate so big jobs can utilize more servers
    
    * get rid of global STATS
    
    * save stats at the end of stage 1 and 2
    
    * create empty stats at the start of stage 2 for jobs that have no stage 1
    
    * download fewer unneeded files from the result folder

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -464,27 +464,30 @@ def download_reference_on_remote(ncbitool_path, input_fasta_ncbi_path, version_n
     execute_command(remote_command(command, key_path, remote_username, server_ip))
     return os.path.join(destination_dir, os.path.basename(input_fasta_ncbi_path))
 
-def download_reference_on_remote_with_version_any_source_type(ref_file, dest_dir,
-    local_ncbitool_dest_dir, remote_ncbitool_dest_dir,
-    key_path, remote_username, server_ip, sudo=False):
+def download_reference_on_remote_with_version_any_source_type(
+        ref_file, dest_dir,
+        local_ncbitool_dest_dir, remote_ncbitool_dest_dir,
+        key_path, remote_username, server_ip, sudo=False):
     # Get input reference and version number.
     # If download does not use ncbitool (e.g. direct s3 link), indicate that there is no versioning.
     input_fasta_name = os.path.basename(ref_file)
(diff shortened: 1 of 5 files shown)


idseq_pipeline/commands/common.py
```diff
@@ -123,14 +123,14 @@ def execute_command(command, progress_file=None):
 
 
 def remote_command(base_command, key_path, remote_username, instance_ip):
-    return 'ssh -o "StrictHostKeyChecking no" -i %s %s@%s "%s"' % (key_path, remote_username, instance_ip, base_command)
+    return 'ssh -o "StrictHostKeyChecking no" -o "ConnectTimeout 15" -i %s %s@%s "%s"' % (key_path, remote_username, instance_ip, base_command)
```
cdebourcy: Is it going to exit gracefully or with an error if timeout is exceeded?

boris-dimitrov: With an error, which would in turn cause the execute_command function to raise an exception.

The default TCP timeout seems to be between 3-4 minutes, and if a machine goes away, this causes the polling for SSH servers to take 3-4 minutes each time that dead machine is polled.   It's better to err out quickly in those cases.  And in general, I've always found it necessary to shorten this timeout in past.

boris-dimitrov: Note this is a timeout just for the handshake to open the TCP connection, not for the command that runs within the SSH session.   There is no timeout for that command.

idseq_pipeline/commands/common.py
```diff
@@ -27,13 +27,72 @@
...
+    def _save(self):
+        with open(self.stats_path, 'wb') as f:
+            json.dump(self.data, f)
+
+    def save_to_s3(self):
+        with self.mutex:
+            self._save()
+            execute_command("aws s3 cp --quiet %s %s/" % (self.stats_path, self.s3_output_dir))
+
+    def get_total_reads(self):
```
boris-dimitrov: Note:  I didn't write most of the code in this class, just pulled it in from already existing functions.

boris-dimitrov:
Validated by this run:   `https://idseq.net/samples/1306`

PR #60, opened 2018-02-21
Accession2seq map

yunfangjuan:
curate the accession_id to file offset database. curated data is already at 

s3://idseq-database/20170824/blast_db/


boris-dimitrov:
Looks good.

idseq_pipeline/commands/curate_accessionid2seq.py
```diff
@@ -0,0 +1,70 @@
...
+        execute_command("mkdir -p %s" % dest_dir)
+        arguments = self.options
+
+        s3_input_file = arguments.get('--input_s3_path')
+        s3_output_file = arguments.get('--output_s3_path')
+        local_input_file = os.path.join(dest_dir, os.path.basename(s3_input_file))
+        local_output_file = os.path.join(dest_dir, os.path.basename(s3_output_file))
+
+        # Copy the data over from s3
+        execute_command("s3mi cp %s %s" % (s3_input_file, local_input_file))
```
boris-dimitrov: Note that s3mi assumes you have at least 24 GB of RAM to spare for buffering.   If your file is bigger than that, and the machine has not enough RAM, s3mi will crash.   If your file is smaller than that, but bigger than the machine RAM, again the machine will crash.   There is an open issue to make s3mi adjust how much RAM it uses based on what's available.

boris-dimitrov: Also btw it's hardly worth using `S3MI` if your destination is a single EBS gp2 volume, since you are steady-state bandwidth limited to 160MB/sec by EBS.

Downloading the STAR genome in host filtering takes >15 minutes without S3MI so making it possible to use S3MI for that could help quite a bit.  We'd have to switch to i3.2xlarge (8vCPU, 61 GB RAM) so we can use NVME for the index download.

boris-dimitrov: Also it looks like STAR is only using 1 CPU core.

yunfangjuan: It's anecdotally a lot faster than a simple s3cp from copynig NT/NR files.

PR #62, opened 2018-02-22
Alignment files generation

yunfangjuan:
Work in progress.
But was able to write the main function for generate the alignment data. It seems to work well. 
Here is an example output file : 
cat nt.species.76773.align_viz.json 

`
{"XM_001731942.1": { "ref_seq": "ATGGGTGTGAATGTGGAAACGATTTCGCCAGGCGATGGCAAGACCTTCCCTAAGGCAGGAGATGTGGTCGACATGCACTACATTGGTACACTGCAGTCCAATGGCAACAAGTTCGACTCTTCGCGTGACCGTGGCCAACCGTTCCGCACAAAGATTGGAGTCGGCCAAGTGATTCGCGGCTGGGACGAGGGTGTGCCCCAGCTCTCGCTTGGTCAAAAGGCAAACCTGATCTGCTCGCCTGACTATGCTTATGGACCTCGCGGATTCCCTCCGAGTACGTGTACAAATTACTGA", 
"reads": [ 
[ "NB501961:60:HFN3JBGX5:1:12307:11356:8573/1", 
 "CCGCTTCAAAGTTGGCCGAGGCGAGGTGATTCGCGGCTGGGACGAAGGAGTTGCCCAGATGTCGGTTGGTCAGAAGGCCAAGCTGATCTGTTCGCCGGAT",  
["84.5", "71", "11", "0", "26", "96", "169", "239", "2.5e-22", "115.0"]
]

]
}
}

`


idseq_pipeline/commands/common.py
```diff
@@ -229,6 +229,21 @@ def percent_str(percent):
     except:
         return str(percent)
 
+def extract_m8_readid(read_id_column):
+    # Get raw read ID without our annotations:
+    # If m8 is from gsnap (case 1), 1 field has been prepended (taxid field).
+    # If m8 is from rapsearch in an old version of the pipeline (case 2), 3 fields have been prepended (taxid, 'NT', NT accession ID).
+    # If m8 is from rapsearch in a newer version of the pipeline (case 3), many fields have been prepended (taxid, alignment info fields),
+    # but the delimiter ":read_id:" marks the beginning of the raw read ID.
```
cdebourcy: maybe we should remove case 3 here as it didn't end up existing?

yunfangjuan: will it be reinstated?

cdebourcy: I don't think so, because I introduced it in a stale PR that tried to do the same alignment stuff you are doing here :)

cdebourcy: Will close now.

cdebourcy: As yours works much better ;)

yunfangjuan: But I do like that format. The current format is kinda like guess work.

idseq_pipeline/commands/accessionid2seq.py
```diff
@@ -9,9 +10,142 @@
...
+        local_json_path = os.path.join(dest_dir, "align_viz")
+
+        # Copy the data over from s3
+        if not local_db_path:
+            local_db_path =  os.path.join(dest_dir, os.path.basename(s3_db_path))
+            if s3_db_path[-3:] == '.gz':
+                local_db_path = local_db_path[:-3]
+                execute_command("aws s3 cp %s - | gunzip > %s" %(s3_db_path, local_db_path))
+            else:
+                execute_command("aws s3 cp %s %s" %(s3_db_path, local_db_path))
```
boris-dimitrov: This entire block you can replace with
```
local_db_path  = fetch_from_s3(s3_db_path, dest_dir, auto_unzip=True)
```
leaving the rest of the code unchanged.   It will correctly skip the download if the file already exists.

idseq_pipeline/commands/common.py
cdebourcy: You could always introduce that format here then! Thus making the documentation true retroactively

idseq_pipeline/commands/accessionid2seq.py
yunfangjuan: I won't use the run function here. I will just use the helper function for stage 2

PR #65, opened 2018-02-27
Alignment backend processing

yunfangjuan:
- Added to post processing step
- Also have a stand alone command to generate the alignment visualization info.

### DIFF
--- a/idseq_pipeline/commands/accessionid2seq.py
+++ b/idseq_pipeline/commands/accessionid2seq.py
@@ -114,7 +114,7 @@ def get_sequence_by_accession_id(accession_id, nt_loc_dict, ntf, nt_bucket, nt_k
             ref_seq = ntf.read(seq_len)
         else:
             # use AWS api
-            ref_seq = nt_key.get_contents_as_string(headers={'Range' : 'bytes=%d-%d' % (range_start, range_start + seq_len) })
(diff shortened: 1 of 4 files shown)


idseq_pipeline/commands/accessionid2seq.py
```diff
@@ -94,7 +125,61 @@ def generate_alignment_viz_json(nt_file, nt_loc_db, db_type,
...
+
+def get_sequence_for_thread(error_flags, accession_info, accession_id, nt_loc_dict, nt_s3_path, semaphore, mutex):
+    try:
+        accession_info['seq'] = get_sequence_by_accession_id(accession_id, nt_loc_dict, None, nt_s3_path)
+    except:
+	mutex.acquire()
+        error_flags["error"] = 1
+        mutex.release()
+    finally:
+        semaphore.release()
```
boris-dimitrov: It's good to also protect with mutex your access to the `accession_info` dict, as it is being written to by multiple threads.   I suggest this.
```
    try:
        result = accession_info['seq'] = get_sequence_by_accession_id(accession_id, nt_loc_dict, None, nt_s3_path)
        with mutex:
            accession_info['seq'] = result
    except:
        with mutex:
            if not error_flags:
                # print the error only from one thread
                traceback.print_exc()
            error_flags["error"] = 1
            return
    finally:
        semaphore.release()
```

yunfangjuan: Cool.  Thanks!

PR #66, opened 2018-03-02
try using s3mi

boris-dimitrov:


boris-dimitrov:
This appears to work well.   It only makes things about 2x faster, because tar and Docker.   If you use the zipped versions, there is hardly any speedup at all (as expected --- and noted in s3mi docs).  But with proper usage it reduces stage 1 duration by about 5 minutes.

PR #70, opened 2018-03-03
make it possible to run pylint and fix linter warnings

boris-dimitrov:
also use lighterweight subprocess.check_output instead of the
real heavyweight execute_command_with_output for the multitude
of s3api get_object range requests


boris-dimitrov:
This considerably improves perf (see issue for discussion), and I am building more on it, so committing as is for now.   Successfully ran on 2 samples.

Issue #71, opened 2018-03-03
generate alignment viz takes a very long time for a small sample


boris-dimitrov:
https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/526890d4-4f8b-4387-a098-ddccd18a502e

https://idseq.net/samples/1309

Only 250k non-host reads

Phase 1 completed in 21 minutes, phase 2 in 35 minutes, phase 3 still going after >1 hour I think.

I will try rerunning it after some fixes I made in https://github.com/chanzuckerberg/idseq-pipeline/pull/70

boris-dimitrov:
```
(4903.49s elapsed)
```

boris-dimitrov:
Each request takes 3-4 seconds, so by issuing 16 concurrent requests we are only getting throughput of 4 requests per second.   I am going to go to 64 concurrent requests and I will skip the deletes when running in a container.

boris-dimitrov:
Apparently this is CPU constrained.   I could only get it to 5 requests/second and that pegs all 4 vCPUs 100% in aws s3api get-object.  I think what we need to do here is run stage3 on the himem queue.

boris-dimitrov:
The python script is also using 71% CPU which means we might not be able to double perf without running python in multiprocessing mode.   It may be that there is a lot of overhead in running "aws s3api" commands for such small amounts of data -- like in opening connections.   So the next thing to try is to coalesce those requests.

boris-dimitrov:
yep, pure CPU --- running stage 3 on the himem queue with 32vCPUs improves throughput from 5 to 28 accession id's per second:   https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/2bcb6bbf-98bc-476c-9baa-03ca32f2c841

boris-dimitrov:
...but eventually slows down to 10 per second

boris-dimitrov:
Another problem is that many of these accessoon IDs are quite large and we are reading them into python and doing a str replace to remove newlines:
```
-rw-r--r-- 1 root root   5545548 Mar  3 09:37 accession-CP011642.1
-rw-r--r-- 1 root root   5578595 Mar  3 09:37 accession-CP011930.1
[...]

PR #73, opened 2018-03-03
get more cpu cores to work on the alignmnet viz

boris-dimitrov:
it seems there are plenty of reference sequences in the 100s of MB, and we typically only need small fragments from those where reads from the sample align;  much faster to only read in those fragments


boris-dimitrov:
This solves https://github.com/chanzuckerberg/idseq-pipeline/issues/71

boris-dimitrov:
Passed test on both short and log reference sequences.   Here is a long example.   https://idseq.net/samples/1309/alignment/nt_genus_9935

PR #76, opened 2018-03-06
move ssh key

cdebourcy:
I have already moved the key to the new location. Will delete it in old location once we push.

Next step:
When no jobs are running (no gsnap/rapsearch machines), delete the key pair in AWS and create a new one with same name. New gsnap/rapsearch machines should then start up with the new key.


boris-dimitrov:
remember to merge this into master only after you have assured that gsnap/rapsearch machines will accept the new key

PR #82, opened 2018-03-13
extract name from accession entry too

yunfangjuan:
So we can show name on the web

### DIFF
--- a/idseq_pipeline/commands/accessionid2seq_functions.py
+++ b/idseq_pipeline/commands/accessionid2seq_functions.py
@@ -194,8 +194,10 @@ def delete_many(files, semaphore=None): #pylint: disable=dangerous-default-value
 def get_sequences_by_accession_list_from_file(accession2seq, nt_loc_dict, nt_file):
     with open(nt_file) as ntf:
         for accession_id, accession_info in accession2seq.iteritems():
-            accession_info['ref_seq'] = get_sequence_by_accession_id_ntf(accession_id, nt_loc_dict, ntf)
-            accession_info['ref_seq_len'] = len(accession_info['ref_seq'])
+            (ref_seq, seq_name) = get_sequence_by_accession_id_ntf(accession_id, nt_loc_dict, ntf)
+            accession_info['ref_seq'] = ref_seq
+            accession_info['ref_seq_len'] = len(ref_seq)
+            accession_info['name'] = seq_name
 
 def get_sequences_by_accession_list_from_s3(accession_id_groups, nt_loc_dict, nt_s3_path):
     threads = []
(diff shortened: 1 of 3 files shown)


idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -234,29 +237,33 @@ def get_sequence_for_thread(error_flags, accession_info, accession_id, nt_loc_di
...
+    return (ref_seq, seq_name)
 
 def get_sequence_by_accession_id_s3(accession_id, nt_loc_dict, nt_bucket, nt_key):
     seq_len = 0
+    seq_name = ''
     entry = nt_loc_dict.get(accession_id)
     if entry:
-        range_start = entry[0]+entry[1]
-        seq_len = entry[2]
+        (range_start, name_length, seq_len) = entry
```
boris-dimitrov: i didn't realize you knew the name length already!   then the simplest solution would have been to simply add the name length to all the seek offsets when you read the accession file, and keep the name at the beginning of the accession file

idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -234,29 +237,33 @@ def get_sequence_for_thread(error_flags, accession_info, accession_id, nt_loc_di
...
         NUM_RETRIES = 3
         for attempt in range(NUM_RETRIES):
             try:
                 pipe_file = 'pipe-{attempt}-accession-{accession_id}'.format(attempt=attempt, accession_id=accession_id)
                 os.mkfifo(pipe_file)
-                get_range = "aws s3api get-object --range bytes=%d-%d --bucket %s --key %s %s" % (range_start, range_start + seq_len - 1, nt_bucket, nt_key, pipe_file)
+                get_range = "aws s3api get-object --range bytes=%d-%d --bucket %s --key %s %s" % (range_start, range_start + name_length + seq_len - 1, nt_bucket, nt_key, pipe_file)
                 get_range_proc = subprocess.Popen(get_range, shell=True, stdout=DEVNULL)
-                subprocess.check_call("tr -d '\n' < {pipe_file} > {accession_file}".format(pipe_file=pipe_file, accession_file=accession_file), shell=True)
+                seq_name = subprocess.check_output("cat {pipe_file} |tee >(tail -n+2 |tr -d '\n' > {accession_file}) |head -1 ".format(pipe_file=pipe_file, accession_file=accession_file), executable='/bin/bash', shell=True).split(" ", 1)[1]
```
boris-dimitrov: This looks nice and clean.    I have some suspicions about `check_output` making it slow -- since that blocks the entire pipeline until the python process is ready to receive -- and was a problem in one of the earlier versions.  But with some luck the buffering in the pipe will be enough to ensure this works fine.

Could you run this on a big job, one that used to take 82 minutes, but now takes 8, and see it is still 8 after your change?

boris-dimitrov:
I do not see where you use `accession_info['name']` --- did you forget to include that code?

idseq_pipeline/commands/accessionid2seq_functions.py
boris-dimitrov: but there is something to be said about the file looking cleaner this way

idseq_pipeline/commands/accessionid2seq_functions.py
yunfangjuan: which sample specifically?

yunfangjuan:
it's there. line 200. in the _functions.py

PR #87, opened 2018-03-21
when a gsnap server hangs it can bring all jobs to a halt as each job will eventually poll it

boris-dimitrov:
not quite ready yet, i cannot believe python doesn't just have a library for this sort of stuff


boris-dimitrov:
This appears to work well here:   https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/1d54cbee-3e74-4f72-b670-3a5e56d42108;filter=chunks

PR #97, opened 2018-03-28
gtf indexing

yunfangjuan:
Add GTF to indexing.


cdebourcy:
Would you mind also adding typical value for GTF file argument into the usage documentation?

PR #98, opened 2018-03-28
turn off lazy run

yunfangjuan:
turn off lazy run for stage 1/2


yunfangjuan:
This is not meant to be merged. Just leave as a branch for option.

boris-dimitrov:
Let's hold off on merging this.   I think lazy run should be turned on for retries of failed jobs, and turned off for reruns of succeeded jobs.   Also it should be turned off if the pipeline version changed since the last run.   Filed issue https://github.com/chanzuckerberg/idseq-pipeline/issues/103

PR #99, opened 2018-03-28
Write a push_reference Python script to replace sh script

jshoe:
- Replacement for push_reference_update.sh to be cleaner in adding the date-getting functionality and also to minimize mixing bash and Python for future changes.
- Does the same thing as push_reference_update.sh except takes out the archive lines since we won't need that anymore + sets the new folder names and new date structure. Gets the max date of the input files in each step.
- Trying to maintain the separation of concerns between this script and the rest of the pipeline and keeping these date changes in this script.
- This isn't ready for use yet w/o changes to the rest of the pipeline.

A few things for the pipeline as a whole we could discuss / keep in mind:
- Having this script outside the main idseq_pipeline.commands module
- The mixing and back-and-forth between bash and Python
- Global module level variables / env variables can get a little confusing?


jshoe:
Going to replace pipeline_command with execute_command..

push_reference_update.py
```diff
@@ -0,0 +1,128 @@
...
+local_work_dir = "idseq_pipeline_temp"
+ncbitool_s3_path = "s3://czbiohub-infectious-disease/ncbitool"
+
+
+def main():
+    ##### PARAMETERS TO EDIT #####
+
+    ## To pull NCBI references through ncbitool, which archives and records versions, set URL_PREFIX=''.
+    # To pull the latest files from NCBI without using ncbitool (no version information), set URL_PREFIX=ftp://ftp.ncbi.nlm.nih.gov.
+    os.environ["URL_PREFIX"] = "ftp://ftp.ncbi.nlm.nih.gov"
```
yunfangjuan: we can make this 
os.environ["URL_PREFIX"] = os.environ.get("URL_PREFIX", "ftp://ftp.ncbi.nlm.nih.gov") 
and ditto for the other env in case we want to overwrite them.

push_reference_update.py
```diff
@@ -0,0 +1,128 @@
...
+    date_setup()
+    make_gsnap_index()
+    make_rapsearch_index()
+    make_lineage_files()
+    make_accession_mapping()
+
+
+def make_gsnap_index():
+    nt = "/blast/db/FASTA/nt.gz"
+    date = set_index_date([nt])
```
yunfangjuan: do we want to set_index_date differently for each of the part of our index. I feel it would be easier to take a snapshot for all the files we need on the same time and use the start of running time for index making

push_reference_update.py
```diff
@@ -0,0 +1,128 @@
...
+    ## Instances with adequate resources for making gsnap and rapsearch indexes.
+    # Need to have write access to destination folders.
+    os.environ["RAPSEARCH_SERVER_IP"] = "54.191.193.210"
+    os.environ["GSNAP_SERVER_IP"] = "34.211.67.166"
+
+    ##### COMMANDS #####
+    date_setup()
+    make_gsnap_index()
+    make_rapsearch_index()
+    make_lineage_files()
```
yunfangjuan: I think the lineage stuff doesn't really need to part of this. But I am fine with it right now. Maybe add a TODO: revisit to decide if we want to make lineage as part of the index making process (or its own)

push_reference_update.py
```diff
@@ -0,0 +1,128 @@
...
+    mapping_files = ["/pub/taxonomy/accession2taxid/nucl_est.accession2taxid.gz",
+                     "/pub/taxonomy/accession2taxid/nucl_gb.accession2taxid.gz",
+                     "/pub/taxonomy/accession2taxid/nucl_gss.accession2taxid.gz",
+                     "/pub/taxonomy/accession2taxid/nucl_wgs.accession2taxid.gz",
+                     "/pub/taxonomy/accession2taxid/pdb.accession2taxid.gz",
+                     "/pub/taxonomy/accession2taxid/prot.accession2taxid.gz"]
+    input_files = [nr] + [nt] + mapping_files
+    date = set_index_date(input_files)
+    pre = os.environ["URL_PREFIX"]
+    dest = "s3://idseq-database/alignment_data/" + date
```
yunfangjuan: avoid hardcoding the destination. Making it an environmental variable with default.

push_reference_update.py
```diff
@@ -0,0 +1,128 @@
...
+    map_env = [pre + f for f in mapping_files]
+    map_env = ','.join(map_env)  # Concatenate into env var
+
+    os.environ["MAPPING_FILES"] = map_env
+    prev_mapping = "s3://czbiohub-infectious-disease/references/accession2taxid.db.gz"
+    cmd = "curate_accession2taxid --mapping_files %s --nr_file %s --nt_file %s --output_s3_folder %s --previous_mapping %s".format(
```
yunfangjuan: idseq_pipeline curate_accession2taxid instead?

push_reference_update.py
jshoe: @yunfangjuan @cdebourcy Ah right so I was thinking to set "Run Time" as 'start of running time for gsnap+rapsearch+etc'. But do you think our "Source Time" should also be a single date across gsnap+rapsearch+etc? Or by snapshot did you mean actually copying the files again?

jshoe:
Hm yes if there are no objections I think I will move this into another subcommand of 'idseq_pipeline'. Not much of a point in keeping it outside the module

push_reference_update.py
yunfangjuan: By snapshoting, I mean the time you get the files. So I think all of them can share the same source time at which we snapshot the data (which is slightly different from when the data was actaully updated at NCBI side.)

jshoe:
Two more things for this (could be in a separate PR):
(1) adding more source files to this script
(2) trivial parallelization of the three blocks (gsnap, rapsearch, accession)

PR #112, opened 2018-03-31
non-paired-end-reads jobs were ALWAYS skipping stage 1

boris-dimitrov:


boris-dimitrov:
sorry about the style here, need to clean that up obviously, but this will have to do for the quick "rapid response" runs tonight

PR #116, opened 2018-04-02
(For Later) Change some env calls to remove defaults

jshoe:
- To be used later!
- Goal is to enforce more separation of concerns so that the pipeline has things set from env vars
- Remove some defaults so that errors are more explicit when they're not provided and we don't get confused when the defaults are used, and the references are changing anyway.
- ACCESSION2TAXID and LINEAGE_SHELF are moved out of common.py to the sub-files in which they are used because otherwise they'd need to be set for every command call or err.
- Let me know if I should go even further or change fewer of them?


idseq_pipeline/commands/blacklist_functions.py
```diff
@@ -6,8 +6,8 @@
 from .common import *
 
 # arguments from environment variables
-INPUT_FASTA_S3 = os.environ.get('INPUT_FASTA_S3')
-ACCESSION2TAXID_DB_S3_PATH =  os.environ.get('ACCESSION2TAXID_DB_S3_PATH')
+INPUT_FASTA_S3 = get_env_or_err('INPUT_FASTA_S3')
```
cdebourcy: FYI 
```
INPUT_FASTA_S3 = os.environ['INPUT_FASTA_S3']
```
would also have the effect of getting the value or error otherwise -- but what you're doing, controlling the error message, is nicer

cdebourcy:
Cool. The gsnap and rapsearch index locations are actually set in idseq-infra: they get downloaded when the gsnap/rapsearch servers are set up. So you may want to raise an accompanying PR for changing those

jshoe:
Note for future:
- Basically the explicitly set paths need to be loaded into the env somehow. E.g. first failure will be on LINEAGE_SHELF being undefined which previously would be set to 's3://czbiohub-infectious-disease/references/taxid-lineages.db'

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -24,11 +24,11 @@
 VERSION_OUT = 'versions.json'
 
 # arguments from environment variables
-INPUT_BUCKET = os.environ.get('INPUT_BUCKET')
-FILE_TYPE = os.environ.get('FILE_TYPE')
-OUTPUT_BUCKET = os.environ.get('OUTPUT_BUCKET')
-STAR_GENOME = os.environ.get('STAR_GENOME', 's3://czbiohub-infectious-disease/references/human/STAR_genome.tar')
-BOWTIE2_GENOME = os.environ.get('BOWTIE2_GENOME', 's3://czbiohub-infectious-disease/references/human/bowtie2_genome.tar')
```
boris-dimitrov: Removing the defaults here also removes some documentation value.   Worth adding a comment to indicate what an example path looks like.

PR #118, opened 2018-04-04
smart lazy run

yunfangjuan:
testing ...


idseq_pipeline/commands/common.py
```diff
@@ -577,6 +578,20 @@ def download_reference_on_remote_with_version_any_source_type(
...
+    try:
+        version_hash = json.loads(execute_command_with_output("aws s3 cp %s - " % version_s3_path))
+        for entry in version_hash:
+            if entry['name'] == 'idseq-pipeline':
+                prev_pipeline_version_sig = re.sub(r'\.\d+$', '', entry['version']);
+                pipeline_version_sig = re.sub(r'\.\d+$', '', pipeline_version)
+                return (prev_pipeline_version_sig != pipeline_version_sig)
+        return True # idseq-pipeline info is not
+    except:
+        # couldn't get the s3_version_file (no output most likely). Return True to indicate change from nothing
```
boris-dimitrov: Please add traceback.print_exc() before returning True here.   That way we'd see in the log what the exception was, just in case it's something unexpected.  (*pun*)

idseq_pipeline/commands/host_filtering.py
```diff
@@ -8,6 +8,7 @@ def run(self):
         from .host_filtering_functions import *
 
         unbuffer_stdout()
+        lazy_run = not big_version_change_from_last_run(self.version, get_host_filtering_version_s3_path())
         upload_commit_sha(self.version)
 
-        run_stage1(True)
+        run_stage1(lazy_run)
```
boris-dimitrov: We have a small problem with this approach here if the pipeline crashes.   It will have uploaded a version string that suggests that any leftover files from the last run are actually from the new version, when in fact they are not.   Is there some way we can delete leftover files as soon as we decide this run shouldn't be lazy.

boris-dimitrov: Alternatively we can put the pipeline version in the path.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -564,8 +567,13 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats):
                       before_filetype="fasta_paired",
                       after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_SAM_OUT1),
                       after_filetype="fasta_paired")
+def init_stage1():
+    global STAR_BOWTIE_VERSION_FILE_S3
+    if not STAR_BOWTIE_VERSION_FILE_S3:
+        STAR_BOWTIE_VERSION_FILE_S3 = get_host_index_version_file()
```
boris-dimitrov: My concern here is that STAR_BOWTIE_VERSION_FILE_S3 is not initialized until this runs.   The better approach is to change the initializer for the global, like so:

STAR_BOWTIE_VERSION_FILE_S3 = os.getenv(STAR_BOWTIE_VERSION_FILE_S3, get_host_index_version_file())

That way any code that refers to this global will access the correctly initialized value.

yunfangjuan: can't do that unless  I put some code before the constants setting. that's why i do init instead.

PR #120, opened 2018-04-04
(Do not merge) new hit calling

cdebourcy:
What this change does:
- Implement new hit calling. Keep up to 1000 hits for each read* and identify the lowest taxonomic rank (species/genus/family/other) at which the hit taxids are all the same. Then produce the same type of JSON that master produces, except master gives the species-level hit for each read (whether that's appropriate or not), whereas my branch gives either a species-level or a genus-level or a family-level hit depending on the analysis. The branch also performs aggregation to levels above the hit level.
- Remove lots of unnecessary and clumsy steps. We now go straight from m8 file to taxon_counts without first making the following intermediate files: annotated m8, annotated fasta, deuterostome filtered fasta, count csv.
- Since intermediate annotated m8 files are removed, we don't need to worry about parsing out the actual read id from the annotated read id anymore when dealing with m8 files.
- Pair concordance is removed because it's too problematic.

To do: 
- Make taxid annotations of fasta (postprocess) consistent with taxon_counts. Right now, they are still in the old style.
- Right now, all the new steps except gsnap & rapsearch are eager. We need to put the lazy logic back in. But since we're thinking about refactoring that logic anyway, we could perhaps do it in a separate diff.
- Potentially, add pair concordance back in.

To load the new results into idseq-web, we'll need to:
- load from `multihit_idseq_web_sample.json` instead of `idseq_web_sample.json`
- remove the logic that keeps only species-level hits then aggregates to get genus count
See https://github.com/chanzuckerberg/idseq-web/pull/1143

*Gsnap and rapsearch have different behaviors when you tell them to keep up to 1000 hits:
- gsnap seems to keep all hits that have the single best e-value
- rapsearch seems to keep all hits that e-values better than the threshold imposed in the command. I've added postprocessing to make it behave the same as gsnap for now.


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -278,6 +291,80 @@ def log_corrupt(is_corrupt, m8_file, line):
...
+        hit_line_columns = hit_line.rstrip("\n").split("\t")
+        read_id = hit_line_columns[0]
+        hit_level = hit_line_columns[1]
+        hit_taxid = hit_line_columns[2]
+        if int(hit_level) < 0:
+            hit_line = hit_f.readline()
+            m8_line = m8_f.readline()
+            continue
+        m8_line_columns = m8_line.split("\t")
+        assert m8_line_columns[0] == hit_line_columns[0], "read_ids in %s and %s do not match: %s vs. %s" % (os.path.basename(m8_file),
```
boris-dimitrov: This assert fired in https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/9de9d30e-6050-4dcd-9e43-7c0f37b9b372

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -278,6 +291,80 @@ def log_corrupt(is_corrupt, m8_file, line):
...
+            e_value = math.log10(e_value)
+        hit_taxid_and_parents = lineage_map.get(hit_taxid, ("-100", "-200", "-300"))[(int(hit_level)-1):3]
+        for i, hit_taxid in enumerate(hit_taxid_and_parents):
+            # TO DO: even if hit is only genus-level, still add a species-level entry for consistency.
+            # The species taxid should be negative to indicate it's not real.
+            # Challenge is to prevent duplicate entries (should be aggregated).
+            # Perhaps a better overall approach would be to record all the hits as before,
+            # but add a flag column to indicate whether the assignment at this level is actually to be trusted
+            # or not (e.g. "mapped to this species ID, but we have determined that the species-level assignment
+            # is meaningless"). Then consume the flag column in the web app to hide/aggregate the meaningless entries.
```
boris-dimitrov: i like the idea of assigning it a pseudo speciesid = 1e9 + abs(genus_taxid), or 1e12 + abs(family_id) in the even the genus mapping is also bogus, and flipping the sign to negative is imo optional but i see the value 

adding flags can easily get out of hand and unless the species id is autogenerated it is arbitrary and therefore not as desirable as a deterministic result

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -712,17 +812,81 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+    read_ids = filter(None, read_ids)
+    read_ids = set(read_ids)
+    sorted_input_m8 = input_m8 + "-sorted"
+    execute_command("sort -k1 %s > %s" % (input_m8, sorted_input_m8))
+    read_to_hit_level = {}
+    read_to_taxid = {}
+    outf = open(output_m8, "wb")
+    outf_sum = open(output_summary, "wb")
+    for read_id in read_ids:
+        # TO DO: remove inefficiency of calling grep on the entire file for each read_id.
```
boris-dimitrov: Yeah this is likely worth addressing, but first see how long it takes for biggish files.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -712,17 +812,81 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+        # Lines with same read_id are contiguous (verify?), so just iterate line by line.
+        m8_lines = execute_command_with_output("grep '^%s' %s" % (read_id, sorted_input_m8)).split("\n")
+        accessions_evalues_lines = [(line.split("\t")[1],
+                                     float(line.split("\t")[10]),
```
boris-dimitrov: calling line.split("\t") multiple times is also an inefficiency

idseq_pipeline/commands/non_host_alignment_functions.py
boris-dimitrov: Actually I'm still debating this with myself...

boris-dimitrov: After in-person discussion in today's 1pm meeting we agreed to start with the singleton pseudogenera approach.

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: oops: `JF317013.1-80 vs. JF317013.1-8`
Need to be more specific with my `grep` and include the trailing tab

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -717,17 +839,81 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+    def call_hit_level(read_id, hits):
+        for i, level in enumerate(["species", "genus", "family"]):
+            taxids = hits[level]
+            taxids = [t for t in taxids if int(t) >= 0] # do not consider unclassified hits. TO DO: decide if that's an improvement
+            n = len(set(taxids)) # number of distinct taxids at this level
+            if n == 1:
+                return i+1, taxids[0]
+        return -1, -1
+    # Deduplicate m8 and summarize hits
+    read_ids = execute_command_with_output("grep -v '^#' %s | cut -f1" % input_m8).split("\n")
```
boris-dimitrov: Just use `subprocess.check_output` here instead of `execute_command_with_output`.   The latter is meant for heavy weight infrequently called commands.

idseq_pipeline/commands/common.py
```diff
@@ -20,6 +20,7 @@
 ACCESSION2TAXID = ("%s/%s/accession2taxid.db.gz" % (base_s3, base_dt))
 base_s3 = bucket + "/taxonomy"
 LINEAGE_SHELF = ("%s/%s/taxid-lineages.db" % (base_s3, base_dt))
+INVALID_CALL_BASE_ID = -1e8 # don't run into -2e9 limit. current largest taxid is around 2e6 so should be fine
```
boris-dimitrov: this -2e9 limit is only when using 32-bit integers yes?   if using 32-bit floating point numbers, the limit is even less (there are only 23 bits of mantissa in fp32... so you start losing precision a lot sooner)

i recommend not using "1e8" notation if your domain is the integers, since "-1e8" is a floating point number;   you can say instead `-(10 ** 8)`

and if you ensure that SQL and rails are both using 64-bit integers for this field, then `-(10*9)` is just fine

cdebourcy: Thanks. It's the limit for the signed integer column in the idseq-web mysql database

idseq_pipeline/commands/common.py
```diff
@@ -623,3 +609,16 @@ def upload_version_tracker(source_file, output_name, reference_version_number, o
 
 def env_set_if_blank(key, value):
     os.environ[key] = os.environ.get(key, value)
+
+def validate_taxid_lineage(taxid_lineage, hit_taxid, hit_level):
+    # Take the taxon lineage and mark meaningless calls with fake taxids.
+    # For each hit, the fake taxids below the meaningful hit level should depend
+    # only on: (a) the taxonomic level (b) the taxid of the meaningful hit
+    cleaned_taxid_lineage = []
+    for i in range(len(taxid_lineage)):
```
boris-dimitrov: If you run pylint on this, it will tell you to replace this line and the next with
```
    for i, taxid in enumerate(taxid_lineage)
```

idseq_pipeline/commands/common.py
```diff
@@ -623,3 +609,16 @@ def upload_version_tracker(source_file, output_name, reference_version_number, o
...
+        taxid = taxid_lineage[i]
+        tax_level = i+1
+        if tax_level < int(hit_level):
```
boris-dimitrov: Here are some minor stylistic suggestions, and you are totally welcome to use them or ignore them;   my feeling is that suggestion 3 may be the more useful one.

1) When expecting an integer value to be passed as a string, you might name the argument `hit_taxid_str` or `hit_level_str` instead of `hit_taxid` and `hit_level`.

2.   Pull out of the loop anything that gets repeated on each iteration.
```
hit_level = int(hit_level_str)
hit_taxid = int(hit_taxid_str)
```

3. Combine the benefits of 1 and 2 above by making it the caller's responsibility to convert hit_taxid and hit_level to integers before passing those arguments into this function.   The arguments can then continue to be called `hit_taxid` and `hit_level`.

In Python3 you could also annotate the function to indicate that those arguments should be integers.   Such annotations in general help correctness and sometimes can help performance.

4.  For a pure function without side effects (that is, a function that does not use or modify any global state, and whose sole purpose is to return a result determined solely from its argument values), you could name it with a noun that describes the value it returns, just like you would a mathematical function;  whereas functions with side effects can be named with a verb.   So this function could be called `cleaned_taxid_lineage`.

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -77,17 +79,32 @@ def accession2taxid(read_id, accession2taxid_dict, hit_type, lineage_map):
     species_taxid, genus_taxid, family_taxid = lineage_map.get(taxid, ("-100", "-200", "-300"))
     return species_taxid, genus_taxid, family_taxid
 
-def generate_taxid_fasta_from_accid(input_fasta_file, accession2taxid_path, lineagePath, output_fasta_file):
+def generate_taxid_fasta_from_accid(input_fasta_file, hit_summary_files, accession2taxid_path, lineagePath, output_fasta_file):
     accession2taxid_dict = shelve.open(accession2taxid_path)
     lineage_map = shelve.open(lineagePath)
+    nt_hit_summary_file = hit_summary_files[0]
+    nr_hit_summary_file = hit_summary_files[1]
```
boris-dimitrov: two long variable names that differ in only one letter -- generally error prone

usually we see this in
```
for my_wonderful_thing in my_wonderful_things:
   do_something(my_wonderful_thing)
```
where we are one typo away from calling `do_something` on the entire list

the shorter the variable name the less problematic this is --- for example `nt` and `nr` are ok since they differ by 50% :P)  though `nT` and `nR` would be better since they emphasize the difference

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -77,17 +79,32 @@ def accession2taxid(read_id, accession2taxid_dict, hit_type, lineage_map):
...
+    def get_valid_hit(read_id, hit_summary_file):
+        hit_summary = subprocess.check_output("grep '^%s\t' %s" % (read_id, hit_summary_file), shell="true").split("\n")
```
boris-dimitrov: Don't you mean `shell=True`?

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -77,17 +79,32 @@ def accession2taxid(read_id, accession2taxid_dict, hit_type, lineage_map):
...
+        assert len(hit_summary) == 1, "More than 1 line for %s in %s" % (read_id, hit_summary_file)
+        hit_line_columns = hit_summary[0].split("\t")
+        hit_level = hit_line_columns[1]
```
boris-dimitrov: I'd say this is the place to convert to int, as it would provide an extra measure of validaton in context where the exception would be easier to debug.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -677,28 +590,39 @@ def interpret_min_column_number_string(min_column_number_string, correct_number_
...
+                         + ['> ' + multihit_remote_outfile, ';'])
+
+    multihit_summary_basename = "summary-" + multihit_basename
+    dedup_multihit_basename = "dedup-" + multihit_basename
+    multihit_summary_file = os.path.join(CHUNKS_RESULT_DIR, multihit_summary_basename)
+    dedup_multihit_local_outfile = os.path.join(CHUNKS_RESULT_DIR, dedup_multihit_basename)
+    is_multihit_present = fetch_lazy_result(os.path.join(SAMPLE_S3_OUTPUT_CHUNKS_PATH, multihit_basename), CHUNKS_RESULT_DIR)
+    is_multihit_summary_present = fetch_lazy_result(os.path.join(SAMPLE_S3_OUTPUT_CHUNKS_PATH, multihit_summary_basename), CHUNKS_RESULT_DIR)
+    is_dedup_multihit_present = fetch_lazy_result(os.path.join(SAMPLE_S3_OUTPUT_CHUNKS_PATH, dedup_multihit_basename), CHUNKS_RESULT_DIR)
+    if not lazy_run or not is_multihit_present or not is_multihit_summary_present or not is_dedup_multihit_present:
```
boris-dimitrov: Since we are always lazy this here makes no difference, but in general, in the eager case, it's an unnecessary penalty in performance to try to fetch_lazy_result if we are then going to overwrite it.

boris-dimitrov: The way this code was written previously, it did not attempt the fetch unless lazy_run=True.

boris-dimitrov: In fact, if any one of these files is missing, there is no benefit to copying the other from s3.

boris-dimitrov: So, you should not break it down on separate lines like this.

Instead, write a little helper function here

```
    def lazy_fetch(basename):
        return fetch_lazy_result(os.path.join(SAMPLE_S3_OUTPUT_CHUNKS_PATH, basename), CHUNK_RESULTS_DIR)
```
then you can write the if condition as follows
```
    if not lazy_run or not all(lazy_fetch(f) for f in [multihit_basename, multihit_summary_basename, dedup_multihit_basename]):
        ...
```

boris-dimitrov: In fact for readability you may want to make to call the helper function `lazy_fetch_all` so it gets a list
```
    def lazy_fetch_all(basenames):
        return all(fetch_lazy_result(os.path.join(SAMPLE_S3_OUTPUT_CHUNKS_PATH, bn), CHUNK_RESULTS_DIR) for bn in basenames)
```
then the if guard would read
```
    if not lazy_run or not lazy_fetch_all([multihit_basename, multihit_summary_basename, dedup_multihit_basename]):
        ...
```

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -709,28 +633,83 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+
     write_to_log("finished alignment for chunk %s" % chunk_id)
-    return os.path.join(CHUNKS_RESULT_DIR, dedup_outfile_basename)
+    return multihit_local_outfile
 
 
+def call_hits_m8(input_m8, output_m8, output_summary):
+    # Get lineage info
+    lineage_path = fetch_reference(LINEAGE_SHELF)
+    lineage_map = shelve.open(lineage_path)
```
boris-dimitrov: This is running concurrently in 64 threads.    Opening the same file from 64 threads doesn't seem good.   But, what does work, according to shelve docs, and based on our experience in accessionid2seq_functions.py, is opening this in the function that spawns all these threads, and passing the dict down as a param all the way through to here.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -709,28 +633,83 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+    accession2taxid_path = fetch_reference(ACCESSION2TAXID)
+    accession2taxid_dict = shelve.open(accession2taxid_path)
```
boris-dimitrov: Same goes for all the shelve databases that you access read-only.

Anything we write to will need locks.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -709,28 +633,83 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+    def call_hit_level(read_id, hits):
+        for i, level in enumerate(["species", "genus", "family"]):
+            taxids = hits[level]
+            taxids = [t for t in taxids if int(t) >= 0] # do not consider unclassified hits. TO DO: decide if that's an improvement
+            n = len(set(taxids)) # number of distinct taxids at this level
+            if n == 1:
+                return i+1, taxids[0]
+        return -1, -1
+    # Deduplicate m8 and summarize hits
+    read_ids = subprocess.check_output("grep -v '^#' %s | cut -f1" % input_m8, shell="true").split("\n")
```
boris-dimitrov: This needs to be `shell=True` not `shell="true"`, correct?

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -709,28 +633,83 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+        first_line = best_accessions_evalues_lines[0][2]
+        outf.write(first_line + "\n")
+
+        hits = { "species": [], "genus": [], "family": [] }
+        for acc in best_accessions:
+            hits = add_taxid_hits(acc, hits)
+        hit_level, taxid = call_hit_level(read_id, hits)
+        outf_sum.write("%s\t%d\t%s\n" % (read_id, hit_level, taxid))
+    outf.close()
+    outf_sum.close()
```
boris-dimitrov: Because this is running inside multiple threads, we don't want to leak open file descriptors if exceptions occur in some of the threads, because good hygiene there would increase the chances of the unaffected chunks to complete.   The classic way of doing that is
```
    f.open()
    try:
        do some work on f
    finally:
        f.close()
```
but python has a thing for this that handles it 
```
with open("...", "") as f:
    do some work on f
```
since for two files we'll have to nest, that means deeper indentation, but that's worth it for the built in exception handling

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -761,24 +740,16 @@ def run_gsnapl_remotely(input_files, lazy_run):
...
+    multihit_chunk_summaries = [f.replace("multihit", "summary-multihit") for f in chunk_output_files]
+    multihit_chunk_dedup_m8s = [f.replace("summary-multihit", "dedup-multihit") for f in multihit_chunk_summaries]
+    for output_item in [(chunk_output_files, MULTIHIT_GSNAPL_OUT),
+                        (multihit_chunk_summaries, SUMMARY_MULTIHIT_GSNAPL_OUT),
+                        (multihit_chunk_dedup_m8s, DEDUP_MULTIHIT_GSNAPL_OUT)]:
+        chunk_files = output_item[0] 
+        concat_basename = output_item[1]
+        concatenate_files(chunk_files, os.path.join(RESULT_DIR, concat_basename))
+        with iostream:
+            execute_command("aws s3 cp --quiet %s/%s %s/" % (RESULT_DIR, concat_basename, SAMPLE_S3_OUTPUT_PATH))
```
boris-dimitrov: Great!

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: yes

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -709,28 +633,83 @@ def run_gsnapl_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work
...
+    sorted_input_m8 = input_m8 + "-sorted"
+    execute_command("sort -k1 %s > %s" % (input_m8, sorted_input_m8))
+    read_to_hit_level = {}
+    read_to_taxid = {}
+    outf = open(output_m8, "wb")
+    outf_sum = open(output_summary, "wb")
+    for read_id in read_ids:
+        # TO DO: remove inefficiency of calling grep on the entire file for each read_id.
+        # Lines with same read_id are contiguous (verify?), so just iterate line by line.
+        m8_lines = subprocess.check_output("grep '^%s\t' %s" % (read_id, sorted_input_m8), shell="true").split("\n")
```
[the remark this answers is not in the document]
cdebourcy: fixed

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: ok

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: fixed

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: ok

idseq_pipeline/commands/non_host_alignment_functions.py
cdebourcy: ok

idseq_pipeline/commands/postprocess_functions.py
cdebourcy: ok

PR #122, opened 2018-04-06
Versioned result folder

yunfangjuan:
Version result folder


idseq_pipeline/commands/common.py
```diff
@@ -583,14 +583,18 @@ def get_host_index_version_file(star_genome):
...
 def big_version_change_from_last_run(pipeline_version, version_s3_path):
     ''' Return True is there's a significant pipeline version chanage. i.e. 1.2.1 -> 1.3.0. False otherwise '''
     try:
         version_hash = json.loads(execute_command_with_output("aws s3 cp %s - " % version_s3_path))
         for entry in version_hash:
             if entry['name'] == 'idseq-pipeline':
-                prev_pipeline_version_sig = re.sub(r'\.\d+$', '', entry['version']);
-                pipeline_version_sig = re.sub(r'\.\d+$', '', pipeline_version)
+                prev_pipeline_version_sig = major_version(entry['version']);
+                pipeline_version_sig = major_version(pipeline_version)
```
boris-dimitrov: Why do we still need function `big_version_change_from_last_run`?

yunfangjuan: we don't need it any more. but keeping it around for future. I can remove it though.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -22,18 +22,21 @@
...
 FILE_TYPE = os.environ.get('FILE_TYPE')
-OUTPUT_BUCKET = os.environ.get('OUTPUT_BUCKET')
+OUTPUT_BUCKET = os.environ.get('OUTPUT_BUCKET').rstrip('/')
 STAR_GENOME = os.environ.get('STAR_GENOME', 's3://czbiohub-infectious-disease/references/human/STAR_genome.tar')
 BOWTIE2_GENOME = os.environ.get('BOWTIE2_GENOME', 's3://czbiohub-infectious-disease/references/human/bowtie2_genome.tar')
 STAR_BOWTIE_VERSION_FILE_S3 = os.environ.get('STAR_BOWTIE_VERSION_FILE_S3', get_host_index_version_file(STAR_GENOME))
 DB_SAMPLE_ID = os.environ['DB_SAMPLE_ID']
 AWS_BATCH_JOB_ID = os.environ.get('AWS_BATCH_JOB_ID', 'local')
+PIPELINE_VERSION = os.environ.get('PIPELINE_VERSION')
+SAMPLE_S3_OUTPUT_POSTFIX = "/%s" % major_version(PIPELINE_VERSION) if PIPELINE_VERSION else ""
```
boris-dimitrov: I think making the version optional is probably more trouble than it's worth.

yunfangjuan: need it for idseq-web compatibility. But I can remove it afterwards.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -564,6 +567,11 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats):
                       after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_SAM_OUT1),
                       after_filetype="fasta_paired")
 
+def upload_pipeline_version_file():
+    execute_command("echo %s > %s" % (PIPELINE_VERSION, PIPELINE_VERSION_OUT))
+    execute_command("aws s3 cp %s %s/" % (PIPELINE_VERSION_OUT, OUTPUT_BUCKET))
```
boris-dimitrov: If OUTPUT_BUCKET does not end in "/" and has not yet been created, the result here is likely not the one intended.

idseq_pipeline/commands/non_host_alignment.py
```diff
@@ -6,9 +6,10 @@ class Non_host_alignment(Base):
 
     def run(self):
         from .non_host_alignment_functions import *
-
         unbuffer_stdout()
-        lazy_run = not big_version_change_from_last_run(self.version, get_alignment_version_s3_path())
         upload_commit_sha(self.version)
+        stage1_version = os.environ.get('HOST_FILTER_PIPELINE_VERSION')
```
boris-dimitrov: The version is actually available in a variable in the running pipeline.  It should not be obtained from the environment like this.

Care should be taken in the docker container init for pipeline stages >1 to checkout a git hash that is exactly the same as the hash that ran in stage 1.

idseq_pipeline/commands/host_filtering_functions.py
yunfangjuan: why? see '/' in the command

idseq_pipeline/commands/non_host_alignment.py
yunfangjuan: it's used for comparison purpose. this command is for alignment. it wouldnt really know how host filtering was run unless informed

idseq_pipeline/commands/common.py
boris-dimitrov: unused code slows down any kind of refactoring so i'm inclined to not keep as a rule

idseq_pipeline/commands/host_filtering_functions.py
boris-dimitrov: i missed that, thanks

PR #123, opened 2018-04-11
Changes for gsnap 2.0 index

jshoe:
- Upgrade to Gsnap 2018
- Accompanying PR for infra: https://github.com/chanzuckerberg/idseq-infra/pull/86
- Infra PR needs to be applied first


cdebourcy:
This is awesome! But I think you will also need to update NT_LOC_DB in postprocess_functions.py

jshoe:
wait where does "nt_loc.db" get generated again?

cdebourcy:
https://github.com/chanzuckerberg/idseq-pipeline/blob/master/idseq_pipeline/commands/accessionid2seq_functions.py

It's relatively new. @yunfangjuan made it for the alignment viz. It needs to go together with the NT that was used for alignment

jshoe:
Ah ok I'll look into this later

cdebourcy:
Cool. @yunfangjuan can let us know if it would perhaps be okay to have a mismatch in the NT and NT_LOC_DB versions for a short interim period

boris-dimitrov:
Oh also hold for @yunfangjuan to comment re the NT_LOC_DB issue.

idseq_pipeline/version/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '1.4.0'
```
boris-dimitrov: Please add a note for version 1.5.0 to the release notes in the top level README.

Here, add a comment reminding the next person who bumps the version to document it in the readme.

idseq_pipeline/version/__init__.py
```diff
@@ -1 +1 @@
...
+__version__ = '1.5.0'
```
cdebourcy: It will now have to be 1.6.0, since Boris's last commit was 1.5.0 :)

PR #125, opened 2018-04-12
key for hit_summary_files should be NT or NR

boris-dimitrov:


boris-dimitrov:
This completed validation on the small sample, see charles https://alpha.idseq.net/samples/508 vs boris https://alpha.idseq.net/samples/509;  identical.

Also produced excellent results on the RR008 tests where previously we were getting bogus spirometra hits;  now it's all filtered out;  just one bacterial species remains with too few reads to be a concern.

PR #128, opened 2018-04-17
keep older webapps from tieing down gsnap/rapsearch for jobs submitte…

boris-dimitrov:
…d with new pipeline


boris-dimitrov:
This passed testing.

PR #130, opened 2018-04-19
fix fasta download

boris-dimitrov:
- 1.6.0
    - Fix fasta downloads broken by release 1.5.0, making sure only
      hits at the correct level are output in the deduped m8.
    - Fix fasta download for samples with unpaired reads by eliminating
      merged fasta for those samples.
    - Extend the partial fix in release 1.5.1 to repair more of the
      broken reports.  Full fix requires rerun with updated webapp.
    - Correctly aggregate counts for species with unclassified genera,
      such as e.g. genus-less species 1768803 from	family 80864.
    - Fix total count in samples with unpaired reads (no longer doubled).
    - Fix crash when zero reads remain after host filtering.
    - Fix bug in enforcing command timeouts that could lead to hangs.
    - Fix performance regression in stage 2 (non-host alignment)
      introduced with 1.5.0.
    - Deduplicate and simplify much of stage 2, and improve performance
      by parallelizing uploads and downloads.


boris-dimitrov:
Manually pushed to master.

PR #131, opened 2018-04-19
fix deuterostome filter

cdebourcy:
move it before aggregation


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -295,6 +298,8 @@ def generate_taxon_count_json_from_m8(m8_file, hit_level_file, e_value_type, cou
 
         # Aggregate each level
         for tax_level, taxid in enumerate(cleaned_hit_taxids_all_levels, 1):
+            if SKIP_DEUTERO_FILTER and taxid in taxids_toremove:
```
boris-dimitrov: Missing NOT?

PR #133, opened 2018-04-21
Assembly prototype

cdebourcy:


idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -11,6 +11,7 @@
 # from common import REF_DIR
 DEST_DIR = ROOT_DIR + '/idseq/data' # generated data go here
 TEMP_DIR = ROOT_DIR + '/tmp' # tmp directory with a lot of space for sorting large files
+SPADES_DIR = ROOT_DIR + '/spades'
```
jshoe: comment for this?

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -318,7 +322,53 @@ def run_stage3(lazy_run=False):
...
+    # run assembly
+    def install_spades():
+        # TODO: update Docker image instead
+        command = "cd %s; " % SPADES_DIR
+        command += "wget http://cab.spbu.ru/files/release3.10.1/SPAdes-3.10.1-Linux.tar.gz; "
+        command += "tar -xzf SPAdes-3.10.1-Linux.tar.gz; "
+        command += "cp SPAdes-3.10.1-Linux/bin/* /usr/local/bin/; "
+        command += "cp -r SPAdes-3.10.1-Linux/share/* /usr/local/share/"
+        subprocess.check_call(command, shell=True)
+    def make_inputs_for_assembly():
```
jshoe: line spacing?

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -318,7 +322,53 @@ def run_stage3(lazy_run=False):
...
+        # Get taxids to sssemble based on criteria from report, currently just species taxids with the most reads
```
jshoe: (a)sssemble?

idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -318,7 +322,45 @@ def run_stage3(lazy_run=False):
...
+            execute_command_realtime_stdout("spades.py -s %s -o %s -m 60 -t 32 --only-assembler" % (input_fasta, tmp_output_dir))
+            subprocess.check_call("mv %s/scaffolds.fasta %s" % (tmp_output_dir, output_fasta), shell=True)
+            return True
+        except:
+            return False
+
+    inputs = make_inputs_for_assembly()
+    for taxid, input_fasta in inputs.iteritems():
+        output_fasta = os.path.join(RESULT_DIR, taxid + ".scaffolds.fasta")
+        if spades(input_fasta, output_fasta):
```
jshoe: If 'spades.py' or 'mv %s/scaffolds.fasta' fail and return False, will that still be logged properly?

idseq_pipeline/commands/assembly_functions.py
```diff
@@ -0,0 +1,134 @@
...
+        taxon_counts = pipeline_output['pipeline_output']['taxon_counts_attributes']
+        sorted_taxids_to_assemble = get_taxids_to_assemble(taxon_counts)
+        # Get reads for those taxids
+        output = {}
+        hit_delimiters = ['_nt', '_nr'] # annotations are like e.g. "genus_nt:543:"
+        for taxid in sorted_taxids_to_assemble:
+            pattern = '\|'.join(['%s:%s:' % (delimiter, taxid) for delimiter in hit_delimiters])
+            partial_fasta =  os.path.join(RESULT_DIR, taxid + ".fasta")
+            try:
+                subprocess.check_call("grep -A 1 --no-group-separator '%s' %s > %s" % (pattern, full_fasta, partial_fasta), shell=True)
```
boris-dimitrov: For efficiency you could build a single grep regular expression, and run grep once.   Not sure how much it matters for this here.  How long does this take for runs with 10k or 20k taxids?

idseq_pipeline/commands/assembly_functions.py
```diff
@@ -0,0 +1,134 @@
...
+        d = {}
+        read_order = []
+        with open(input_fasta, 'rb') as f:
+            for line in f:
+                if line.startswith('>'):
+                    read_name = line
+                    read_order += [read_name]
+                else:
+                    d[read_name] = d.get(read_name, '') + line
+        return d, read_order
```
boris-dimitrov: Some fasta files are 30GB uncompressed.   How large are the files this is meant to work for?

idseq_pipeline/commands/assembly_functions.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def max_read_length(input_fasta):
+        fasta_dict, _read_order = read_fasta(input_fasta)
+        return max([length_without_newlines(sequence) for sequence in fasta_dict.values()])
+
+    def spades(input_fasta, output_fasta):
+        tmp_output_dir = input_fasta + "_temp_output"
+        tmp_output_file = tmp_output_dir + "/scaffolds.fasta"
+        try:
+            execute_command_realtime_stdout("spades.py -s %s -o %s -m 60 -t 32 --only-assembler" % (input_fasta, tmp_output_dir))
```
boris-dimitrov: nb:  I am removing `execute_command_realtime_stdout`.  Just use `execute_command`.

idseq_pipeline/commands/assembly_functions.py
```diff
@@ -0,0 +1,134 @@
...
+            subprocess.check_call("mv %s %s" % (tmp_output_file, output_fasta), shell=True)
+            return True
+        except:
+            return False
```
boris-dimitrov: When you swallow an exception it's usually good to print the stacktrace in the log, i.e., `traceback.print_exc()`.

boris-dimitrov: It probably won't be super informative in this case.

boris-dimitrov:
One concern for me is if these tasks end up taking a very long time on the idseq-himem queue, they will prevent jobs from starting stage 1.   May be a good idea to give the assembler its own queue, 'idseq-assembly' and back that by a the standard compute environment or one that has like 1 guaranteed machine and room to grow.

Another question I have about this, is there a concept of lazyness, so if it crashes can you rerun so it will resume from where it crashed?

cdebourcy:
Ok, I will create a new queue with lower priority than stage 1. I’m also planning to benchmark how long fastas of various sizes take to assemble and just skip assembly for anything that could potentially take too long, for now.

idseq_pipeline/commands/assembly_functions.py
cdebourcy: This is used only on the annotated non-host fasta, which contains at most 2M reads (due to subsampling), corresponding to about 650 MB in practice. Still, I will see if I can get rid of read_order somehow

idseq_pipeline/commands/assembly_functions.py
cdebourcy: Is `execute_command` still going to give realtime stdout then? For SPAdes it's quite useful as it is telling you exactly what it is doing on the reads

idseq_pipeline/commands/assembly_functions.py
cdebourcy: ok!

idseq_pipeline/commands/assembly_functions.py
cdebourcy: Ok, I rewrote it so that it doesn't read the file into memory

PR #140, opened 2018-04-28
speed up hit calling when accession ids repeat a lot

boris-dimitrov:
we see very slow hit calling after rapsearch with tons of iops
and very little cpu, consistent with looking up a lot of entries
in the accession2taxid dict out of order, in examples where
there are ~75 hits per read


boris-dimitrov:
Tested on sample 1006 in prod.

Issue #144, opened 2018-04-30
math domain error in a fraction of hit calling jobs even with older gsnap


boris-dimitrov:
at first i thought this only happened with the new gsnap but seems to occasionally happen with the old gsnap so thus probably a bug in the hit calling code

Exception: Failed generate_taxon_count_json_from_m8 on ['/mnt/idseq/data/idseq-samples-production-samples-48-3280-results-1.6/results/dedup.multihit.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.m8', '/mnt/idseq/data/idseq-samples-production-samples-48-3280-results-1.6/results/summary.multihit.gsnapl.unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.tab', 'raw', 'NT', '/mnt/idseq/ref/taxid-lineages.db', '/mnt/idseq/ref/deuterostome_taxids.txt', 100385936, 407066, '/mnt/idseq/data/idseq-samples-production-samples-48-3280-results-1.6/results/nt_multihit_idseq_web_sample.json', False]

File "/mnt/idseq-pipeline/idseq_pipeline/commands/non_host_alignment_functions.py", line 329, in generate_taxon_count_json_from_m8
e_value = math.log10(e_value)
ValueError: math domain error

cc @jshoe @cdebourcy 

https://idseq.net/samples/3280/pipeline_runs/

boris-dimitrov:
The problem here appears to be a bug in GSNAP.    For the following two reads
```
>A00111:139:HCYVYDMXX:2:2423:2139:5165/1
GACCACGACCTGGCCGGTGGTGCCGTCCGCCGTGGTGACCGTGAAGGTCTCGGTCAGCGTCTTGCCGGCGCCCAGGCCCTGGACGGCCGCATCGCCGTTGTTCAGCTGGTAGCTCCACTTGCCGTCGGCATCGATGGAGAAGCTGCCATG
>A00111:139:HCYVYDMXX:2:2423:2139:5165/2
GCCCCAGACCAATGCAGCGGGCCAGCATGGCAGCTTCTCCATCGATGCCGACGGCAAGTGGAGCTACCAGCTGAACAACGGCGATGCGGCCGTCCAGGGCCTGGGCGCCGGCAAGACGCTGACCGAGACCTTCACGGTCACCACGGCGGA
```
the output of gsnap is
```
[...]

PR #146, opened 2018-04-30
handle some corrupt m8 lines without crashing

boris-dimitrov:
https://github.com/chanzuckerberg/idseq-pipeline/issues/144


boris-dimitrov:
Fix confirmed.

PR #149, opened 2018-04-30
WIP - curate NT based on sequence length

cdebourcy:


cdebourcy:
Revealed that number of records affected was negligible

PR #158, opened 2018-05-02
Closes #150: do not remove artificial species ids

### LINKED ISSUE #150: Alignment data not showing for non-species specific reads
Example: 
https://idseq.net/samples/3387

This showed up after the hit calling change.

cdebourcy:


yunfangjuan:
Lgtm

PR #161, opened 2018-05-02
Closes #151: make run_in_subprocess a decorator

### LINKED ISSUE #151: run_in_subprocess decorator
Convert this pattern to a python decorator:

```
def generate_taxon_count_json_from_m8(m8_file, hit_level_file, e_value_type, count_type, lineage_map_path, deuterostome_path, total_reads, remaining_reads, output_file, fork=True):
    # TODO: Make this pattern a decorator, ensuring thus-decorated functions always run in subrpocess.
    if fork:
        fork = False
        run_in_subprocess(
            target=generate_taxon_count_json_from_m8,
            args=[m8_file, hit_level_file, e_value_type, count_type, lineage_map_path, deuterostome_path, total_reads, remaining_reads, output_file, fork]
        )
        return
```

cdebourcy:


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -115,6 +115,17 @@ def result_dir(basename):
 MISSING_GENUS_ID = -200
 MISSING_FAMILY_ID = -300
 
+# decorators
+def run_in_subprocess(target):
+    """ Forks a job function to a subprocess """
```
boris-dimitrov: May want to add some docs here on when it is appropriate to decorate a function with this, and any special considerations, for example:

1)   Use this when the function uses a lot of CPU, so it gets to run on its own CPU core.

2)   Typical subprocess caveats apply:

          -- The function may read the values of global variables but when it writes to globals the changes won't be seen by the parent process.   So special constructs from multiprocessing module, or operating system files, pipes, etc, must be used to pass information between parent and child.

            -- As a special case, any use of global threading semaphores, locks, etc, is mute.   Use multiprocessing locks, semaphore, etc instead.

     3)  Performance caveats.   While this helps with CPU, it doesn't help with IO.   If you are planning to run multiple subprocesses that all read the same data file, with a lot of seeking, you should consider finding a way to do all the IO before the subprocesses are launched so that the data is in memory and available to all of them without I/O.   Otherwise they will each do the same I/O, possibly hitting OS bottlenecks.

Also check what's the return value for this sort of invocation.  Probably it should be none & ignored, but I'm not super clear on that.

boris-dimitrov: Point 3 is actually kind of complicated.  Not sure how much is helpful to put in here without incurring the risk of saying something that's not always true.  Use your judgement for what might be most surprising/interesting or the worst caveats to try to protect a naive programmer from.

boris-dimitrov: may be here is a great example use:

```
@run_in_subprocess
def process_fasta(fasta_file):
     do something expensive on fasta_file and write result to fasta_file.new

# Do NOT do this
for f in [r1.fasta, r2.fasta, merged.fasta]:
    process_fasta(f)

# Do this
threads = []
for f in [r1.fasta, r2.fasta, merged.fasta]:
    t = threading.Thread(target=process_fasta, args=[f])
    t.start()
    threads.append(t)
for t in threads:
    t.join()
```
In other words, don't forget this is just a way to cover for the fact that python threads run on the same CPU core so we are only using this decorator to let the threads use more cores.

cdebourcy: Thanks! I'll ad some concise documentation.
The decorator "run_in_subprocess" returns the function "wrapper" -- so the function decorated using "@run_in_subprocess" will return whatever "wrapper" returns (in this particular case, nothing).

boris-dimitrov: Yeah, my point is what wrapper returns is not in any way related to what the target function returns, so this doesn't offer a clean way of passing a result.    Anyway this is good enough for us for now;  I am sure there is some "future" based pattern that just does this right.  You can google for "futures" later.  I hate futures. :)

cdebourcy: I see. I think the "multiprocessing.Pool" is what people use to handle return values

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -643,44 +634,34 @@ def run_chunk(part_suffix, remote_home_dir, remote_index_dir, remote_work_dir, r
         multihit_remote_outfile=multihit_remote_outfile if service == "gsnap" else multihit_remote_outfile[:-3]  # strip the .m8 for rapsearch as it adds that
     )
     if not lazy_run or not fetch_lazy_result(multihit_s3_outfile, multihit_local_outfile):
-        correct_number_of_output_columns = 12
-        min_column_number = 0
-        max_tries = 2
-        try_number = 1
-        while min_column_number != correct_number_of_output_columns and try_number <= max_tries:
+        @retry(2)
```
boris-dimitrov: Let's not apply the decorator here.   Keep it in mind for situations where it's easier to use.

boris-dimitrov:
Charles, I shamelessly stole this, made a couple small changes, tested, and merged into master.

PR #163, opened 2018-05-03
WIP: Gsnap filter

yunfangjuan:
still testing. hope this works!!


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -39,11 +39,15 @@
 ROOT_DIR = '/mnt'
 DEST_DIR = ROOT_DIR + '/idseq/data' # generated data go here
 REF_DIR = ROOT_DIR + '/idseq/ref' # referene genome / ref databases go here
-
+# input files
+EXTRACT_UNMAPPED_FROM_BOWTIE_SAM_OUT1 = 'unmapped.bowtie2.lzw.cdhitdup.priceseqfilter.unmapped.star.1.fasta'
```
boris-dimitrov: Why is this duplicated in two different files; not related to your change, but seems like this stuff belongs into common.py (for now, until we make a special file for it).

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -504,6 +548,11 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats):
                       after_filename=os.path.join(RESULT_DIR, STAR_OUT1),
                       after_filetype=initial_file_type_for_log)
 
+    # start downloading gsnap genome after STAR step
+    local_gsnap_genome = fetch_from_s3(GSNAP_GENOME,
+                                       os.path.join(REF_DIR, os.path.basename(GSNAP_GENOME)),
+                                       auto_unzip=False)
```
boris-dimitrov: i think this wasn't using s3mi because you didn't tell it to allow_s3mi=True

but in any case you shouldn't just copy a tar file from s3 (by whatever means) and then untar it, because that just doubles the amount of data that needs to be written to disk, which is the actual slow part

May I suggest you simply use the command `s3mi cat <s3_path> | tar xvf - -C /mnt/idseq/ref` or if you don't like s3mi just do `aws s3 cp <s3_path> - | tar xvf - -C /mnt/idseq/ref` and start that in a thread at the beginning.   By the time its result is needed, it would have finished.

idseq_pipeline/commands/non_host_alignment_functions.py
yunfangjuan: i dont understand either. i was very confused with it.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -854,18 +858,33 @@ def fetch_input_and_replace_whitespace(input_filename, result):
...
+def get_input_file_list():
+    # Check existence of gsnap filter output
+    try:
+        output = execute_command_with_output("aws s3 ls %s/%s" % (SAMPLE_S3_INPUT_PATH, EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT1))
+        if len(output) > 10: # gsnap filter applied
+            return [EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT1,
+                    EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT2,
+                    EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT3]
+    except:
+        write_to_log("gsnap filter not applied for host filtering")
```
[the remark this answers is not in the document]
yunfangjuan: great

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -471,18 +481,52 @@ def run_bowtie2(input_fas):
...
+
+    # Run Gsnap
+    gsnap_params = [GSNAPL,
+                    '-A sam',
+                    '--batch=0',
+                    '--use-shared-memory=0',
+                    '--gmap-mode=all',
+                    '--npaths=1',
+                    '--ordered',
+                    '-t 32',
```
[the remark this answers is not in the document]
yunfangjuan: ok

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -566,8 +614,22 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats):
     stats.count_reads("run_bowtie2",
                       before_filename=os.path.join(RESULT_DIR, LZW_OUT1),
                       before_filetype="fasta_paired",
-                      after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_SAM_OUT1),
+                      after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_BOWTIE_SAM_OUT1),
```
boris-dimitrov: Please note the function get_remaining_reads() in common.py.   You will have to update it so it now picks up the reads after gsnap not the reads after bowtie2.
```
    def get_remaining_reads(self):
        return (item for item in self.data if item.get("task") == "run_bowtie2").next().get("reads_after")
```

boris-dimitrov: Or whichever is available, I guess.

idseq_pipeline/commands/host_filtering_functions.py
yunfangjuan: OK. I was confused about how fetch_from_s3 work

idseq_pipeline/commands/host_filtering_functions.py
yunfangjuan: OK. interesting. why does it need it in pipeline?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -471,18 +481,52 @@ def run_bowtie2(input_fas):
...
+    else:
+        generate_unmapped_singles_from_sam(RESULT_DIR + '/' + GSNAP_FILTER_SAM, output_prefix)
+    write_to_log("extracted unmapped fragments from SAM file for gsnap output")
+    execute_command("aws s3 cp --quiet %s/%s %s/;" % (RESULT_DIR, GSNAP_FILTER_SAM, SAMPLE_S3_OUTPUT_PATH))
+    execute_command("aws s3 cp --quiet %s/%s %s/;" % (RESULT_DIR, EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT1, SAMPLE_S3_OUTPUT_PATH))
+    if len(input_fas) == 2:
+        execute_command("aws s3 cp --quiet %s/%s %s/;" % (RESULT_DIR, EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT2, SAMPLE_S3_OUTPUT_PATH))
+        execute_command("aws s3 cp --quiet %s/%s %s/;" % (RESULT_DIR, EXTRACT_UNMAPPED_FROM_GSNAP_SAM_OUT3, SAMPLE_S3_OUTPUT_PATH))
 
 def get_host_filtering_version_s3_path():
```
boris-dimitrov: I don't see a caller for this function, can you explain?

idseq_pipeline/commands/common.py
```diff
@@ -95,6 +97,11 @@ def count_reads(self, func_name, before_filename, before_filetype, after_filenam
             self.data = new_data
         self.data.append({'task': func_name, 'reads_before': records_before, 'reads_after': records_after})
 
+    def set_remaining_reads(self):
+        last_entry = self.data[-1] or {}
+        remaining = last_entry.get('reads_after', 0)
+        self.data.append({'remaining_reads': remaining})
```
boris-dimitrov: If here we append, we should search backgward in get_remaining_reads above :)  in case multiple updates

however unlikely

minor point

don't care :)

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -566,8 +609,24 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats):
     stats.count_reads("run_bowtie2",
                       before_filename=os.path.join(RESULT_DIR, LZW_OUT1),
                       before_filetype="fasta_paired",
-                      after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_SAM_OUT1),
+                      after_filename=os.path.join(RESULT_DIR, EXTRACT_UNMAPPED_FROM_BOWTIE_SAM_OUT1),
                       after_filetype="fasta_paired")
+    # run gsnap against host genomes (only available for human as of 5/1/2018)
+    if check_s3_file_presence(GSNAP_GENOME):
```
boris-dimitrov: What is GSNAP_GENOME going to be for mosquito jobs?   Will it exist?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -352,8 +362,10 @@ def run_star(fastq_files):
     def unmapped_files_in(some_dir):
         return ["%s/Unmapped.out.mate%d" % (some_dir, i+1) for i in range(num_fastqs)]
     genome_dir = fetch_genome(STAR_GENOME)
-    # If we are here, we are also going to need a bowtie genome later;  start fetching it now
+    # If we are here, we are also going to need a bowtie genome and gsnap genome later;  start fetching it now
     threading.Thread(target=fetch_genome, args=[BOWTIE2_GENOME]).start()
+    if check_s3_file_presence(GSNAP_GENOME):
```
boris-dimitrov: Could you launch this in Bowtie2?   Where it does
```
fetch_genome(bowtie2)
```
change it to
```
fetch_genome(bowtie2)
if check_s3_file_presence(GSNAP_GENOME):
    threading.Thread(target=fetch_genome, args=[GSNAP_GENOME]).start()
```
The reason is that s3mi is pretty good at utilizing the machine resources and you don't want to have 2 semis running in parallel with star.   Doing as I suggest will ensure only 1 semi at a time.  And it's fine to have it fetch while bowtie is running.

PR #167, opened 2018-05-04
(WIP) Parallelize / asynchronize some AWS cp calls

jshoe:
- execute_command is blocking so for the appropriate "aws s3 cp" calls (mostly uploads) we can use a thread handler pattern instead.

- "with iostream" wraps the call in a semaphore / basically just limits the number of concurrent IO calls currently running.

- How does this look so far? If it's reasonable I'll change more calls in other files.

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -665,8 +665,7 @@ class AsyncHandler:
         pass
 
     def launch(self, target, args):
-        with iostream:  # semaphore to limit max number of operations
-            t = threading.Thread(target=target, args=args)
+        t = threading.Thread(target=target, args=args)
         self.threads.append(t)
         t.start()
 
@@ -675,7 +674,10 @@ class AsyncHandler:
             t.join()
 
     def awsUpload(self, local, remote):
-        self.launch(execute_command("aws s3 cp --quiet %s %s" % (local, remote)), None)
+        self.launch(execute_command, "aws s3 cp --quiet %s %s" % (local, remote))
+
+    def launchCommand(self, command):
+        self.launch(execute_command, command)
 
 
 async_handler = AsyncHandler()
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -211,6 +211,41 @@ class ProgressFile(object):
             self.tail_subproc.kill()
 
 
+class AsyncHandler:
+    def __init__(self):
+        self.threads = []
+        pass
+
+    def launch(self, target, args):
+        if type(args) == str:
+            args = (args, )
+        t = threading.Thread(target=target, args=args)
+        self.threads.append(t)
+        t.start()
+
+    def wait_on_all(self):
+        for t in self.threads:
+            t.join()
+
+    def launch_aws_cp(self, src, dst):
+        self.launch(self.aws_cp_work, (src, dst))
+
+    def aws_cp_work(self, src, dst):
+        with iostream:
+            execute_command("aws s3 cp --quiet %s %s" % (src, dst))
+
+    def launch_command(self, command):
+        self.launch(execute_command, command)
+
+    def launch_io_command(self, command):
+        self.launch(self.io_work, command)
+
+    def io_work(self, command):
+        with iostream:
+            execute_command(command)
+
+async_handler = AsyncHandler()
+
 def execute_command_with_output(command, progress_file=None, timeout=None, grace_period=None):
     with CommandTracker() as ct:
         with print_lock:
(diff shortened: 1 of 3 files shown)


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -828,7 +823,7 @@ def run_generate_unidentified_fasta(input_fa, output_fa):
     #TODO  remove annotated fasta intermediate file and replace > with : below
     subprocess.check_output("grep -A 1 '>NR::NT::' %s | sed '/^--$/d' > %s" % (input_fa, output_fa), shell=True)
     write_to_log("finished job")
-    execute_command("aws s3 cp --quiet %s %s/" % (output_fa, SAMPLE_S3_OUTPUT_PATH))
+    async_handler.launch_aws_cp(output_fa, SAMPLE_S3_OUTPUT_PATH + "/")
```
boris-dimitrov: We'll get in trouble when we start decorating these functions with @run_in_subprocess if they upload something to s3 at the very end.   If you are adding async handler inside such a function, better to pull the upload outside of the function.   Doesn't really apply to this specific instance, but something to think about.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -1071,7 +1052,9 @@ def run_additional_steps():
         execute_command("echo This file is deprecated since pipeline version 1.5.  Please use {new} instead. > {deprecated}".format(
             deprecated=result_dir(DEPRECATED_BOOBYTRAPPED_COMBINED_JSON_OUT),
             new=MULTIHIT_COMBINED_JSON_OUT))
-        execute_command("aws s3 cp --quiet %s/%s %s/" % (RESULT_DIR, DEPRECATED_BOOBYTRAPPED_COMBINED_JSON_OUT, SAMPLE_S3_OUTPUT_PATH))
+        src = RESULT_DIR + "/" + DEPRECATED_BOOBYTRAPPED_COMBINED_JSON_OUT
```
boris-dimitrov: We can actually delete this soon, as the comment suggests.

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -1125,9 +1108,7 @@ def thread_succeeded(name):
     t_annotation.join()
     assert thread_succeeded("annotation")
 
-    for thread_name, t in uploader_threads.items():
-        t.join()
-        assert thread_succeeded(thread_name), "thread {} failed".format(thread_name)
+    async_handler.wait_on_all()
```
boris-dimitrov: You are just being polite here.  I like it.   But if you skipped this, it would happen by default.  The main program would just wait for all threads.

idseq_pipeline/commands/common.py
```diff
@@ -211,6 +211,43 @@ def __exit__(self, *args):
...
+        self.threads.append(t)
+        t.start()
+
+    def wait_on_all(self):
+        write_to_log("Waiting on AsyncHandler threads...")
+        for t in self.threads:
+            t.join()
+        write_to_log("AsyncHandler threads finished.")
+
+    def launch_aws_cp(self, src, dst):
```
boris-dimitrov: I would prefer if you call this `lauch_aws_upload()` to indicate we are only really using it for uploads.

idseq_pipeline/commands/common.py
```diff
@@ -211,6 +211,43 @@ def __exit__(self, *args):
...
 
 
+class AsyncHandler:
+    def __init__(self):
+        self.threads = []
+        pass
+
+    def launch(self, target, args):
+        if type(args) == str:
+            args = (args, )
```
boris-dimitrov: Let's keep this to straight args, matching the threading and multiprocessing interfaces, and more streamlined.

idseq_pipeline/commands/common.py
```diff
@@ -211,6 +211,43 @@ def __exit__(self, *args):
...
+            execute_command("aws s3 cp --quiet %s %s" % (src, dst))
+
+    def launch_command(self, command):
+        self.launch(execute_command, command)
+
+    def launch_io_command(self, command):
+        self.launch(self.io_work, command)
+
+    def io_work(self, command):
+        with iostream:
```
boris-dimitrov: make this a processing semaphore

boris-dimitrov:
Now that i've reflected on this more, I think we only need the joining functionality if we are checking the result status of the uploads.   And this code never checks the result status.  Let's have it do that.  May be have the join function raise an exception if something didn't go well with some upload.

boris-dimitrov:
Looks like a test run for this hung stage 2.   https://github.com/chanzuckerberg/idseq-pipeline/issues/171

idseq_pipeline/commands/non_host_alignment_functions.py
jshoe: wait should I delete this whole block?

jshoe:
@boris-dimitrov would you recommend adding back a "results = {}", "results[run_name] = 'blah'" sort of pattern to support checking the run status?

Or perhaps something like:

class MyThread(Thread):
    def run(self):
        try:
            Thread.run(self)
        except Exception as self.err:
            pass # or raise
        else:
            self.err = None

jshoe:
- May roll this into Boris's uploader changes since it's similar but saving the AsyncHandler class for usage.
- Main conclusion from the test runs is to be careful about doing it in run_in_subprocess.
- Also for some reason the stage is not marked as completed once you change the semaphore and the 'thread' to multiprocessing.semaphore and multiprocessing.process. Some values are not being changed/shared properly?

PR #168, opened 2018-05-04
port retry decorator from autoscaling and add run_in_subprocess decorator

boris-dimitrov:


idseq_pipeline/commands/common.py
```diff
@@ -218,6 +219,70 @@ def __exit__(self, *args):
...
+        # errors won't perturb other random streams used in the application.
+        @wraps(operation)
+        def wrapped_operation(*args, **kwargs):
+            remaining_attempts = attempts
+            delay = base_delay
+            while remaining_attempts > 1:
+                try:
+                    return operation(*args, **kwargs)
+                except:
+                    time.sleep(delay * (1.0 + randgen()))
```
jshoe: I'm sure there's a reason but what's the point of adding the random fractional delay?

boris-dimitrov:
Tested here:  https://idseq.net/samples/3415/
Seems to work.

idseq_pipeline/commands/common.py
boris-dimitrov: To avoid collisions from multiple clients calling the same API endpoint.   Helpful with rate-limited APIs like AWS.

boris-dimitrov: @jshoe You may like this blog post:  https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

PR #172, opened 2018-05-05
fix errors

boris-dimitrov:
i had assumed the retry decorator had been copied directly from
autoscaling.py where it was throroughly tested, but it seems
some errors had been introduced, and also an omission of the
functool wrapper in the run_in_process decorator

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -252,35 +252,33 @@ def run_in_subprocess(target):
     decorated function, do the I/O before the first call, to avoid accessing
     the file multiple times.
     """
+    @wraps(target)
     def wrapper(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/common.py
```diff
@@ -252,35 +252,33 @@ def compute_something(x, y, z):
...
-                    return operation(*args, **kwargs)
-                except:
-                    time.sleep(delay * (1.0 + randgen()))
-                    delay *= exponential_backoff_delay_multiplier
-                    remaining_attempts -= 1
-            # The last attempt is outside try/catch so caller can handle exception
-            return operation(*args, **kwargs)
-        return wrapped_operation
-    return decorator
+def retry(operation, randgen=random.Random().random):
```
boris-dimitrov: Direct paste form autoscaling, where this had been well tested;   the version I copied from the other branch had some changes that made it break.

Issue #169, opened 2018-05-04
timeout enforcement bug


boris-dimitrov:
"Command 227 postprocessing after 135.0 seconds."

This shouldn't happen.

Happens here:   https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logEventViewer:group=/aws/batch/job;stream=aegea_batch/default/59b2ad0e-f24d-4b42-8f3c-68bc868e4d14;filter=Command%20227

cc @jshoe

boris-dimitrov:
Looks like a race condition between timer thread cancel and timer thread relaunch.   Need a lock.

```
class Updater(object):

    def __init__(self, update_period, update_function):
        self.update_period = update_period
        self.update_function = update_function
        self.timer_thread = None
        self.t_start = time.time()

    def relaunch(self, initial_launch=False):
        if self.timer_thread and not initial_launch:
            t_elapsed = time.time() - self.t_start
            self.update_function(t_elapsed)
        self.timer_thread = threading.Timer(self.update_period, self.relaunch)
        self.timer_thread.start()

    def __enter__(self):
        self.relaunch(initial_launch=True)
        return self

    def __exit__(self, *args):
        self.timer_thread.cancel()
```

boris-dimitrov:
It's probably mostly harmless aside from leaking threads and polluting the log with print statements.

boris-dimitrov:
Fixed in #174

Issue #173, opened 2018-05-05
do not use S3 file presence as indicator of whether gsnap should be run


boris-dimitrov:
the check_s3_file_presence function does not distinguish operational errors from legitimate absence of the file in s3;  i researched what work would be needed to distinguish operational error from file absence and it's nontrivial;  the approach of doing retries is just a bandaid

so instead why don't we have a flag that specifies whether or not gsnap should run;   if the flag says to run gsnap, and we can't fetch the index in a couple retries, the job should just crash

right now if we can't fetch the index we just skip gsnap and steam ahead without flagging any error

there is a similar problem where stage 2 determines whether or not it is running on single or paired end reads by just checking how many files it can see in s3;  again this is not the most robust way to handle it;  the metadata is available in the UI so should be passed to the pipeline;  the pipeline should thus know what to expect, and assert if the state of s3 mismatches the expectation

boris-dimitrov:
Btw I already fixed the "similar problem where stage 2 determines whether or not it is running on single or paired end reads by just checking how many files it can see in s3".    It now looks to see if stage 1 ran gsnap by checking the entry in status.json.

PR #174, opened 2018-05-05
misc cleanup and more robust support for multiprocessing

boris-dimitrov:


idseq_pipeline/commands/common.py
```diff
@@ -37,7 +37,7 @@
 
 # peak network & storage perf for a typical small instance is saturated by just a few concurrent streams
 MAX_CONCURRENT_COPY_OPERATIONS = 8
-iostream = threading.Semaphore(MAX_CONCURRENT_COPY_OPERATIONS)
+iostream = multiprocessing.Semaphore(MAX_CONCURRENT_COPY_OPERATIONS)
```
boris-dimitrov: As we use more multiprocessing, threading semaphores are no longer reliable.

idseq_pipeline/commands/common.py
```diff
@@ -71,23 +71,19 @@ def save_to_s3(self):
             self._save()
             execute_command("aws s3 cp --quiet %s %s/" % (self.stats_path, self.s3_output_dir))
 
-    def get_total_reads(self):
-        for item in self.data:
-            if "total_reads" in item:
-                return item["total_reads"]
-        # check run star if total reads not available
+    def get_item(self, function_name):
```
boris-dimitrov: get_total_reads is no longer so special;   if we have a predigested post-filtered input, that came with a "total_reads";  otherwise we initialize "total_reads" at the start of host filtering;   so it's always there, and these old gymnastics are unnecessary

idseq_pipeline/commands/common.py
```diff
@@ -130,9 +126,12 @@ def __init__(self, update_period, update_function):
         self.update_period = update_period
         self.update_function = update_function
         self.timer_thread = None
+        self.exited = False
         self.t_start = time.time()
 
     def relaunch(self, initial_launch=False):
+        if self.exited:
```
boris-dimitrov: This fixes https://github.com/chanzuckerberg/idseq-pipeline/issues/169

idseq_pipeline/commands/common.py
```diff
@@ -145,12 +144,13 @@ def __enter__(self):
...
         self.timer_thread.cancel()
+        self.exited = True
 
 
 class CommandTracker(Updater):
 
-    lock = threading.RLock()
-    count = 0
+    lock = multiprocessing.RLock()
+    count = multiprocessing.Value('i', 0)
```
boris-dimitrov: This counter needs to be incremented from multiple processes when they all execute commands.   This requires it to be wrapped in a special container, called multiprocessing.Value.

idseq_pipeline/commands/common.py
```diff
@@ -281,57 +280,38 @@ def wrapped_operation(*args, **kwargs):
     return wrapped_operation
 
 
-def execute_command_with_output(command, progress_file=None, timeout=None, grace_period=None):
+def execute_command(command, progress_file=None, timeout=None, grace_period=None, with_output=False):
```
boris-dimitrov: Previously we only supported timeout enforcement for execute_command_with_output.   Now we also support it for execute_command.

idseq_pipeline/commands/common.py
```diff
@@ -281,57 +280,38 @@ def wrapped_operation(*args, **kwargs):
...
-    execute_command_realtime_stdout(command, progress_file)
+def execute_command_with_output(command, progress_file=None, timeout=None, grace_period=None):
+    return execute_command(command, progress_file, timeout, grace_period, with_output=True)
 
 
 def remote_command(base_command, key_path, remote_username, instance_ip):
     return 'ssh -o "StrictHostKeyChecking no" -o "ConnectTimeout 15" -i %s %s@%s "%s"' % (key_path, remote_username, instance_ip, base_command)
 
 
-def check_s3_file_presence(s3_path):
```
boris-dimitrov: As the comment here suggests, this test is totally unreliable.   Doing it right with the s3 head-object operation requires careful testing of error codes, and still seems brittle.   The bottom line is that you can never be certain a file doesn't exist in s3, you can only increase your belief by retrying, which has a cost.   So it's best to avoid using this function --- and luckily, we don't actually need it.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -240,30 +240,83 @@ def uncompressed(s3genome):
...
 
 def fetch_genome(s3genome, mutex=threading.RLock(), mutexes={}): #pylint: disable=dangerous-default-value
+    """Fetch and expand genome archive from s3 into local dir.  Return that local dir.
+    If all retries fail, return None.  If already downloaded, return right away.
+    If a fetch of the same genome is already in progress on another thread, wait for it.
+
+    Typical use:
+
+            fruitfly_dir = fetch_genome("s3://fruitfly_genome.tar")
+            assert fruitfly_dir != None
```
boris-dimitrov: We can get rid of this assert business as soon as we address issue https://github.com/chanzuckerberg/idseq-pipeline/issues/173

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -829,34 +829,25 @@ def fetch_input_and_replace_whitespace(input_filename, result):
     s3_input_path = os.path.join(SAMPLE_S3_INPUT_PATH, input_filename)
     s3_output_path = os.path.join(SAMPLE_S3_OUTPUT_PATH, input_filename) # usually the same as the input path
     cleaned_input_path = result_dir("nospace.%s" % input_filename)
-    if not check_s3_file_presence(s3_input_path):
-        result[0] = None
-    else:
+    try:
```
boris-dimitrov: Similarly to issue #173 here, we should examine the stats from stage 1 to determine how many files we are dealing with and what to expect;  instead of trusting that if we don't see a file in S3 it isn't meant to be there.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -598,103 +603,159 @@ def run_gsnap_filter(input_fas):
...
+            assert fn.endswith(".fasta") or fn.endswidth(".fasta.gz"), "Prefiltered input is not a fasta file: {fn}".format(fn=fn)
+            if fn.endswidth(".fasta"):
+                execute_command("mv {fn} {bto}".format(fn=fn, bto=btos[i]))
+            else:
+                t = MyThread(target=unzip_to_file, args=[fn, btos[i]])
+                t.start()
+                unzippers.append(t)
+        for t in unzippers:
+            t.join()
+            assert t.completed and not t.exception
```
boris-dimitrov: cc @jshoe

PR #176, opened 2018-05-08
truncate input if excessive

cdebourcy:
Goes with https://github.com/chanzuckerberg/idseq-web/pull/1266

### DIFF
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -75,6 +75,8 @@ BOWTIE2 = "bowtie2"
 
 # pipeline configuration
 LZW_FRACTION_CUTOFF = 0.45
+MAX_INPUT_LINES = 3*10**8 # must be multiple of 4 (fastq)
+INPUT_TRUNCATED_FILE = 'input_truncated_to_%s' % str(MAX_INPUT_LINES)
 
 # convenience functions
 def fq2fa(input_fastq, output_fasta):
@@ -597,6 +599,14 @@ def run_stage1(lazy_run=True):
     if len(fastq_files) not in [1, 2]:
         write_to_log("Number of input files was neither 1 nor 2. Aborting computation.")
         return # only support either 1 file or 2 (paired) files
+
+    # Truncate files if they are too long
+    for f in fastq_files:
+        initial_size = os.path.getsize(f)
+        execute_command("sed -i '%s,$ d' %s" % (str(MAX_INPUT_LINES + 1), f))
+        final_size = os.path.getsize(f)
+    if final_size < initial_size:
+        execute_command("echo '' | aws s3 cp - %s/%s" % (SAMPLE_S3_OUTPUT_PATH, INPUT_TRUNCATED_FILE))
 
     # Record total number of input reads
     initial_file_type_for_log = "fastq" if "fastq" in FILE_TYPE else "fasta"
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -76,7 +76,7 @@ BOWTIE2 = "bowtie2"
 # pipeline configuration
 LZW_FRACTION_CUTOFF = 0.45
 MAX_INPUT_LINES = 3*10**8 # must be multiple of 4 (fastq)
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -75,6 +75,8 @@
 
 # pipeline configuration
 LZW_FRACTION_CUTOFF = 0.45
+MAX_INPUT_LINES = 3*10**8 # must be multiple of 4 (fastq)
```
jshoe: maybe " 3*(10**8) "

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -598,6 +600,14 @@ def run_stage1(lazy_run=True):
         write_to_log("Number of input files was neither 1 nor 2. Aborting computation.")
         return # only support either 1 file or 2 (paired) files
 
+    # Truncate files if they are too long
+    for f in fastq_files:
+        initial_size = os.path.getsize(f)
+        execute_command("sed -i '%s,$ d' %s" % (str(MAX_INPUT_LINES + 1), f))
+        final_size = os.path.getsize(f)
+    if final_size < initial_size:
```
jshoe: this should be in the 'for' block right?

idseq_pipeline/commands/host_filtering_functions.py
cdebourcy: that does look nicer

idseq_pipeline/commands/host_filtering_functions.py
cdebourcy: Actually no - if there are multiple fastq_files, it should still only upload one "input_truncated" file (number of lines in all fastq_files is expected to be the same). Good point though, it would be better to make the assumption of equal line counts explicit using an assert statement.

idseq_pipeline/commands/host_filtering_functions.py
boris-dimitrov: I've seen this more customarily written as `30*1000*1000`.

idseq_pipeline/commands/host_filtering_functions.py
boris-dimitrov: Yeah let's do some input validation as part of the streaming download and uncompress.   I had added that in my pending PR but reverted due to nontrivial error handling implications;   so will defer for future.

boris-dimitrov: This is how I am changing this in my pending commit, partly to address Jonathan's concern, but also changing the truncation to be to number of reads (which is 1/2 number of lines for fasta and 1/4 for fastq).   Note this number is multipled by the number of files as per our usual convention of counting a paired end read as 2 reads.  Is that useful, Charles, or should I keep it number of lines?
```
    # Truncate files if they are too long
    initial_sizes = []
    for f in fastq_files:
        initial_sizes.append(os.path.getsize(f))
        execute_command("sed -i '%s,$ d' %s" % (str(MAX_INPUT_LINES + 1), f))
        final_sizes.append(os.path.getsize(f))
    if any(fsz < isz for fsz, isz in zip(final_sizes, initial_sizes)):
        execute_command("echo %d | aws s3 cp - %s/%s" % (MAX_INPUT_READS * len(fastq_files), SAMPLE_S3_OUTPUT_PATH, INPUT_TRUNCATED_FILE))

```

cdebourcy: @boris-dimitrov This is nice in that it is able to handle single vs paired inputs. But your comment made me realized that fastas are not handled correctly in either your implementation or mine. The fastq format always has 4 lines per record. The fasta format allows the the sequence to be split over multiple lines, so the records could potentially have variable numbers of lines.

boris-dimitrov: Yikes.   Is it true the only fastas are from Joe's pipeline?

boris-dimitrov: Seems like in the fasta case we should truncate the file on the position of the N+1'st '>' character?   We can create an issue for that.

boris-dimitrov: I am marking it with TODO in the code.

cdebourcy: Yes. Thanks

boris-dimitrov:
Folded in #174.

PR #182, opened 2018-05-15
make truncating input simpler by limiting the STAR input

yunfangjuan:
- Need to discuss about parallelizing total read counts.

### DIFF
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -827,12 +827,13 @@ def run_stage1(lazy_run=True):
         write_to_log("Post-filtered input with {total_reads} original total reads.".format(total_reads=total_reads))
     except:
         stats_in = None
-        total_reads = count_reads(fastq_files[0], initial_file_type_for_log)
         write_to_log("Unfiltered input with {total_reads} total reads.".format(total_reads=total_reads))
 
     # Record total number of input reads
     stats = StatsFile(STATS_OUT, RESULT_DIR, None, SAMPLE_S3_OUTPUT_PATH)
-    stats.data.append({'total_reads': total_reads})
+
+    #total_reads = count_reads(fastq_files[0], initial_file_type_for_log)
+    #stats.data.append({'total_reads': total_reads})
 
     # run host filtering
     run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats, stats_in != None)
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -700,6 +700,11 @@ def run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats, 
              "output_version_file": os.path.join(RESULT_DIR, VERSION_OUT)})
         run_and_log_s3(logparams, target_outputs["run_star"], lazy_run, SAMPLE_S3_OUTPUT_PATH,
                        run_star, fastq_files, uploader_start, total_counts_from_star)
+        if not total_counts_from_star.get('total_reads'):
+            # Total reads not set. Most likely it's lazy run. Damn it. would have to actually count the reads.
+            # TODO: Remove this one we also lazy load the stats.json file
+            total_counts_from_star['total_reads'] = count_reads(fastq_files[0], initial_file_type_for_log)
+
         stats.data.append(total_counts_from_star)
         stats.count_reads("run_star",
                           before_filename=fastq_files[0],
@@ -857,7 +862,7 @@ def run_stage1(lazy_run=True):
         write_to_log("Post-filtered input with {total_reads} original total reads.".format(total_reads=total_reads))
     except:
         stats_in = None
-        write_to_log("Unfiltered input with {total_reads} total reads.".format(total_reads=total_reads))
+        write_to_log("Unfiltered input. Need host filtering")
 
     # run host filtering
     run_host_filtering(fastq_files, initial_file_type_for_log, lazy_run, stats, stats_in != None)
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -211,6 +211,13 @@ def generate_unmapped_pairs_from_sam(sam_file, output_prefix):
                 with open(sam_file, 'rb') as samf:
                     generate_unmapped_pairs_from_sam_work(output_read_1, output_read_2, output_merged_read, samf)
 
+def max_input_lines(input_file):
+    ''' Returning number of lines corresponding to MAX_INPUT_READS based on file type '''
+    if "fasta" in input_file:
+        return MAX_INPUT_READS * 2
+    # assume it's fastq
+    return MAX_INPUT_READS * 4
+
 
 # job functions
 def run_star_part(output_dir, genome_dir, fastq_files, count_genes=False):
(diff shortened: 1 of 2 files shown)


idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -227,12 +227,17 @@ def run_star_part(output_dir, genome_dir, fastq_files, count_genes=False):
                            '--genomeDir', genome_dir,
                            '--readFilesIn', " ".join(fastq_files)]
     if fastq_files[0][-3:] == '.gz':
-        star_command_params += ['--readFilesCommand', 'zcat']
+        if "fasta" in fastq_files[0]:
```
jameshowardwang: these checks are strictly less precise than the old code - e.g. a file named "thisisnotafastafile.fastq.gz" would meet this condition. also "thisisnotafastqfile.fa.gz" would be considered a fastq file

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -227,12 +227,17 @@ def run_star_part(output_dir, genome_dir, fastq_files, count_genes=False):
...
+            max_lines = MAX_INPUT_READS * 2
+        else:
+            max_lines = MAX_INPUT_READS * 4
+        # create a custom decompressor which does "zcat $input_file | head - ${max_lines}"
+        execute_command("echo 'zcat ${2} | head -${1}' > %s/gzhead; " % genome_dir)
```
jameshowardwang: sorry i don't understand where ${2} and ${1} are coming from - are those in the bash environment?

idseq_pipeline/commands/host_filtering_functions.py
yunfangjuan: STAR takes both fasta/fastq, it would figure it out itself we are just checking if it's gzipped or not.

idseq_pipeline/commands/host_filtering_functions.py
yunfangjuan: it's generating a shell script. ${2} is supplied by star as the input file. ${1} is the number of max lines. See documentation above the line.

yunfangjuan:
the max_lines are set by us (which is x4 for fastq). The original implementation with sed is really not performant which transforms the content back and forth and is basically doing head -{4x#reads}

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -821,25 +808,11 @@ def truncate_input(fastq_file):
         assert t.completed and not t.exception
 
     # Downloaded files here are unzipped but not yet truncated.
```
boris-dimitrov: Looks like this comment is no longer accurate after your change.   Perhaps remove the comment.

boris-dimitrov:
Just an aside, it may be ideal to also employ this piping technique for streaming the output of STAR through the pair synchornizer -- so we don't have to wait an hour for it to synchronize the pairs -- it can do that while STAR is slowly outputting them.

An alternative method that would also work is to chunk the STAR inputs.   Even if STAR runs the chunks serially, the pair synchronizer can work on the STAR output for each chunk in parallel.

boris-dimitrov:
So far as approving this pull request, as is, the main thing it's missing is the INPUT_TRUNCATED file.   It is also producing the wrong total count, by counting the untruncated input.

boris-dimitrov:
Re @jameshowardwang's concern about splitting a read in the middle, this ought not to happen for fastq files, where each read is guaranteed to occupy 4 lines;   it might happen for fasta inputs, and we have filed it away as an issue to may be consider, but deprioritized as we don't see that in our inputs so far.

idseq_pipeline/commands/host_filtering_functions.py
jameshowardwang: i know - it's a minor thing but 'if "fasta" in...' just checks if 'fasta' is anywhere in the filename, while the old code looked for it as an extension

idseq_pipeline/commands/host_filtering_functions.py
jameshowardwang: ah i see. would it make sense to just check in gzhead somewhere and call it rather than recreating it each time?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -670,18 +692,32 @@ def uploader_check_wait_all():
...
         logparams = return_merged_dict(
             DEFAULT_LOGPARAMS,
             {"title": "STAR",
              "version_file_s3": STAR_BOWTIE_VERSION_FILE_S3,
              "output_version_file": os.path.join(RESULT_DIR, VERSION_OUT)})
-        run_and_log_s3(logparams, target_outputs["run_star"], lazy_run, SAMPLE_S3_OUTPUT_PATH, run_star, fastq_files, uploader_start)
+        run_and_log_s3(logparams, target_outputs["run_star"], lazy_run, SAMPLE_S3_OUTPUT_PATH,
+                       run_star, fastq_files, uploader_start, total_counts_from_star)
+        if not total_counts_from_star.get('total_reads'):
+            # Total reads not set. Most likely it's lazy run. Damn it. would have to actually count the reads.
```
boris-dimitrov: Aptly put...

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -670,18 +692,32 @@ def uploader_check_wait_all():
...
+            # TODO: Remove this one we also lazy load the stats.json file
+            total_counts_from_star['total_reads'] = count_reads(fastq_files[0], initial_file_type_for_log)
+
+        stats.data.append(total_counts_from_star)
```
boris-dimitrov: Does it matter that total_counts_from_star may have a "truncated" key?   I am guessing not.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -670,18 +692,32 @@ def uploader_check_wait_all():
...
         stats.count_reads("run_star",
                           before_filename=fastq_files[0],
                           before_filetype=initial_file_type_for_log,
                           after_filename=os.path.join(RESULT_DIR, STAR_OUT1),
                           after_filetype=initial_file_type_for_log)
 
+        if total_counts_from_star.get('truncated'):
```
boris-dimitrov: There is a small subtlety here.    If this is a lazy rerun of something that was truncated, `total_counts_from_star` here wouldn't have the `truncated` key.   However, presumably, the `INPUT_TRUNCATED_FILE` would have been uploaded during the previous (non-lazy) run of STAR.   So this is undetectable from the webapp's point of view, but still I'd say it's a tricky scenario.

boris-dimitrov: In case someone in future adds logic the branches based on whether input was truncated or not, you might want to make the check for that here, not in the run_star function, so it can handle the lazyness.

And guess what?   When we get rid of run_and_log_S3, the run_star function will make its own lazyness determination, and will be able to handle this truncation signalling protocol perfectly.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -820,46 +838,31 @@ def truncate_input(fastq_file):
...
+
     # Download total_reads.json input, if present.  This is only provided with post-filtered inputs,
     # where we don't have the reads prior to host filtering.
+    # Record total number of input reads
     try:
         stats_in = StatsFile(STATS_IN, RESULT_DIR, SAMPLE_S3_INPUT_PATH, SAMPLE_S3_OUTPUT_PATH)
         stats_in.load_from_s3()
         total_reads = stats_in.get_total_reads()
         assert total_reads == int(total_reads)
+        stats.data.append({'total_reads': total_reads})
```
boris-dimitrov: Technically this append needs to be outside this try/catch block.   The way this try/catch block is setup is such that if an exception is thrown that means there was no stats_in file provided.  Now you are adding another statement to it that can throw an exception for some unrelated reason, which would then erroneously be suppressed and treated as though no stats_in has been provided.  So this code change, which appears like a minor style change, is actually a semantic change.   The correct behavior when there is an exception in stats.data.append is to dump a stacktrace and crash the program, instead of silently continue pretending on stats in was given.

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -820,47 +842,34 @@ def truncate_input(fastq_file):
...
         stats_in = None
-        total_reads = count_reads(fastq_files[0], initial_file_type_for_log)
-        write_to_log("Unfiltered input with {total_reads} total reads.".format(total_reads=total_reads))
-
-    # Record total number of input reads
-    stats = StatsFile(STATS_OUT, RESULT_DIR, None, SAMPLE_S3_OUTPUT_PATH)
-    stats.data.append({'total_reads': total_reads})
+        write_to_log("Unfiltered input. Need host filtering")
 
+    if total_reads: # set total reads if available
```
boris-dimitrov: `if total_reads == None` would be better, because 0 is a valid total_reads but evaluates to False as an if guard.

PR #186, opened 2018-05-16
pr template and list of samples to test

yunfangjuan:
request template


boris-dimitrov:
cool these are all big samples, pls note that you may want to start testing with smaller toy examples first;   leave these for last

PR #202, opened 2018-05-17
Subset of changes from Python 2to3 tool

jshoe:
- Subset of changes from Python 2to3 tool.
- Everything here should be compatible with either python2 or python3, except for some more memory overhead on python2 for now (iterators being rendered).
- Gets us closer to switching to python3 soon.

- Print with parens on Python2: https://stackoverflow.com/questions/6182964/why-is-parenthesis-in-print-voluntary-in-python-2-7
- Items vs iteritems in python2: https://stackoverflow.com/questions/10458437/what-is-the-difference-between-dict-items-and-dict-iteritems
- Range vs xrange: https://www.geeksforgeeks.org/range-vs-xrange-python/

- Test run: https://staging.idseq.net/samples/554


jshoe:
Actually it looks like some of these print changes aren't needed. Interesting that it changed them

idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -151,9 +151,9 @@ def generate_alignment_viz_json(nt_file, nt_loc_db, db_type,
     result_dict = {}
     to_be_deleted = []
     error_count = 0
-    for accession_id, ad in groups.iteritems():
+    for accession_id, ad in groups.items():
```
cdebourcy: Maybe we should bundle the change from iteritems to items with the change from python2 to actually using python3? So we don't increase the memory overhead

jshoe: Hm ok I'll change this PR to just the print statements

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -169,7 +169,7 @@ def subsample_single_fasta(input_files_basename, target_n_reads, randgen=random.
     write_to_log("target reads: %d" % target_n_reads)
     if total_records <= target_n_reads:
         return input_file
-    records_to_keep = set(randgen.sample(xrange(total_records), target_n_reads))
```
cdebourcy: same comment here, let's make the change from xrange to range once we actually switch to python3. Subsample code is already pretty heavy-weight.

idseq_pipeline/commands/accessionid2seq_functions.py
cdebourcy: Sounds good. Another option could be to check if we're using python2 or python3 and choose iteritems / items as a consequence?

Issue #194, opened 2018-05-17
create separate iostream_uploads semaphore


boris-dimitrov:
when the number of concurrent uploads reaches 8, we stall the entire pipeline, because we can no longer download inputs for the next (ready) step due to the iostreams semaphore being maxed out;   so the next step is delayed until a couple uploads complete;   which may take several minutes

this problem is actually quite noticeable on stool samples with 70+ mil non host reads such as sample 4000

the solution is simple;  create another semaphore iostream_uploads with value 4, to limit the number of concurrent uploads to 4, then for each upload job replace this
```
with iostream:
    aws s3 cp localfile s3://path/remotefile
```
with this
```
with iostream_uploads:
    with iostream:
        aws s3 cp localfile s3://path/remotefile
```
Please make sure to acquire them in this order, not the opposite.

Note this functionality is reasonably well encapsulated in stage 1 (host filtering) but much more rudimentary and haphazard in stage 2, so fix in stage 1 and then port it so it works the same way in both stages.

@cdebourcy @yunfangjuan @jameshowardwang

PR #203, opened 2018-05-17
Add another upload semaphore

jshoe:
For #194 

- Basically create another upload semaphore to reserve some capacity for downloads or file moves to proceed. Interesting approach.

We have:
```
MAX_CONCURRENT_COPY_OPERATIONS = 8
iostream = multiprocessing.Semaphore(MAX_CONCURRENT_COPY_OPERATIONS)
# Make a second semaphore for uploads to reserve some capacity for downloads.
MAX_CONCURRENT_UPLOAD_OPERATIONS = 4
iostream_uploads = multiprocessing.Semaphore(MAX_CONCURRENT_UPLOAD_OPERATIONS)
```

Problem:
"when the number of concurrent uploads reaches 8, we stall the entire pipeline, because we can no longer download inputs for the next (ready) step due to the iostreams semaphore being maxed out; so the next step is delayed until a couple uploads complete; which may take several minutes"

This PR: Replaced:
```
with iostream:
```
with:
```
with iostream_uploads:
    with iostream:
```

- The diff looks a little bigger b/c git didn't match the nested if blocks that well.

- Test run for this change: https://staging.idseq.net/samples/550

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -498,23 +498,12 @@ def fetch_from_s3(source,
             # It's okay if the parent directory already exists, but all other errors are fatal.
             if e.errno != os.errno.EEXIST:
                 raise
-        with iostream_uploads:  # Limit concurrent uploads so as not to stall the pipeline.
-            with iostream:      # Still counts toward the general semaphore.
-                try:
-                    if allow_s3mi:
-                        try:
-                            install_s3mi()
-                        except:
-                            allow_s3mi = False
-                    if unzip:
-                        pipe_filter = "| gzip -dc "
-                    else:
-                        pipe_filter = ""
+        with iostream:
+            try:
+                if allow_s3mi:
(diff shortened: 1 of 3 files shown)


yunfangjuan:
looks reasonable.

idseq_pipeline/commands/common.py
```diff
@@ -491,36 +494,37 @@ def fetch_from_s3(source,
             # It's okay if the parent directory already exists, but all other errors are fatal.
             if e.errno != os.errno.EEXIST:
                 raise
-        with iostream:
-            try:
-                if allow_s3mi:
+        with iostream_uploads:  # Limit concurrent uploads so as not to stall the pipeline.
```
boris-dimitrov: Not here.   This function does downloads, not uploads.

jshoe: fixed thanks

PR #208, opened 2018-05-23
Assorted style changes

jshoe:
- I renamed some things for clarity/generality esp when the function does more than the name suggested. Ex: unbuffer_stdout() unbuffers stdout but also redirects stderr into stdout so I named it set_up_stdout().
- Should be a 1:1 refactoring without any actual functional changes (except maybe print statements) and without anything too fancy.
- Plan is to put up a couple of PRs like these gradually.


idseq_pipeline/commands/common.py
```diff
@@ -662,10 +664,11 @@ def upload_commit_sha(version, s3_destination=None):
 
     # also initialize the main version output file with the commit sha and the job_id
     global OUTPUT_VERSIONS
-    aws_batch_job_id = os.environ.get('AWS_BATCH_JOB_ID', 'local')
```
cdebourcy: Did you remove this line by accident?

jshoe: @cdebourcy no it's the same as line 657 || 659

jshoe:
@jameshowardwang Test run for this (in progress) https://staging.idseq.net/samples/561

jshoe:
@jameshowardwang Test run finished https://staging.idseq.net/samples/561?pipeline_version=1.7&background_id=4

PR #216, opened 2018-05-26
WIP - status file for result monitor

cdebourcy:
Uploads a status file that lets the web app know when a stage is complete.
Goes with https://github.com/chanzuckerberg/idseq-web/pull/1318


boris-dimitrov:
You may want to bump the minor-most version number for a change like this.

PR #217, opened 2018-05-30
(Minor) Changes to address PR comments

jshoe:
- Changed strings back to something like "".format() so that the linters will be able to catch mismatched string interpolation params.
- Nice commentary input from Boris and Charles

- Test run for this change: https://staging.idseq.net/samples/586

- I plan on running bigger/more test runs every few changes.


idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -1165,6 +1224,8 @@ def run_stage2(lazy_run=True):
                                  AWS_BATCH_JOB_ID)
     configure_logger(log_file)
 
+    write_to_log("Starting stage...")
```
jameshowardwang: where was this going before?

jshoe: Hah I added it last week but it didn't do anything if above configure_logger

PR #220, opened 2018-05-31
(Minor) More style changes (mostly to common.py)

jshoe:
- Mostly minor style changes and documentation
- Some more logging print statements
- Tried to reduce some verbosity

- Test run for this: https://staging.idseq.net/samples/595


idseq_pipeline/commands/postprocess_functions.py
```diff
@@ -316,29 +317,33 @@ def run_stage3(lazy_run=False):
                         (SAMPLE_S3_INPUT_PATH, os.path.basename(local_file),
                          INPUT_DIR))
 
-    # generate taxid fasta
-    log_params = return_merged_dict(
-        DEFAULT_LOG_PARAMS,
-        {"title": "run_generate_taxid_fasta_from_hit_summaries"})
+    def default_log_param_plus_title(title):
+        return return_merged_dict(DEFAULT_LOG_PARAMS, {"title": title})
```
jameshowardwang: since this is the only place DEFAULT_LOG_PARAMS is used, maybe we can get rid of the global and just put that hash in here?

idseq_pipeline/commands/common.py
```diff
@@ -601,8 +640,8 @@ def run_and_log_work(log_params, target_outputs, lazy_run, func_name,
     global OUTPUT_VERSIONS
     global run_and_log_mutex
 
-    LOGGER = logging.getLogger()
-    LOGGER.info("========== %s ==========", log_params.get("title"))
+    logger = logging.getLogger()
```
jameshowardwang: why isn't this using write_to_log?

PR #222, opened 2018-06-01
(Minor) More style changes for host filtering functions

jshoe:
- Most of the diff is adding a "result_path" function to do `os.path.join(RESULT_DIR, basename)`
- Test run for this PR: https://staging.idseq.net/samples/622
- Will bump the version number before merging

### DIFF
--- a/idseq_pipeline/commands/host_filtering_functions.py
+++ b/idseq_pipeline/commands/host_filtering_functions.py
@@ -345,7 +345,7 @@ def fetch_genome_work(s3genome, strict):
             install_s3mi()
             tarfile = uncompressed(s3genome)
             try:
-                print("Trying to download compressed genome...")
+                write_to_log("Trying to download compressed genome...")
                 cmd = "s3mi cat {tarfile} | tar xvf - -C {refdir}".format(
                     tarfile=tarfile, refdir=REF_DIR)
                 execute_command(cmd)
@@ -363,7 +363,7 @@ def fetch_genome_work(s3genome, strict):
                     assert os.path.isdir(genome_dir)
                 else:
                     # Okay, may be s3mi is broken.  We'll try aws cp next.
-                    print("Error in downloading with s3mi. Trying aws cp...")
+                    write_to_log("Error in downloading with s3mi. Trying aws cp...")
                     raise
         except:
             try:
@@ -610,7 +610,7 @@ def run_star(fastq_files, uploader_start, total_counts_from_star):
             uploader_start(output_i, SAMPLE_S3_OUTPUT_PATH + "/")
     # Cleanup
     execute_command("cd %s; rm -rf *" % SCRATCH_DIR)
-    print("Finished running STAR.")
+    write_to_log("Finished running STAR.")
 
 
 def run_priceseqfilter(input_fqs, uploader_start):
@@ -706,8 +706,8 @@ def run_lzw(input_fas, uploader_start):
 
 
 def run_bowtie2(input_fas, uploader_start):
-    """Bowtie is an aligner used for filtering out host genomes. Two input
-    FASTAs means paired reads.
+    """Bowtie2 is an aligner we use for filtering out reads that map to the
+    host genome. Two input FASTAs means paired reads.
 
     http://bowtie-bio.sourceforge.net/index.shtml
     """
@@ -744,7 +744,7 @@ def run_bowtie2(input_fas, uploader_start):
     else:
         bowtie2_params.extend(['-U', input_fas[0]])
     execute_command(" ".join(bowtie2_params))
-    print("Finished Bowtie alignment.")
+    write_to_log("Finished Bowtie alignment.")
 
     # Extract out unmapped files from sam
     output_prefix = result_path(EXTRACT_UNMAPPED_FROM_BOWTIE_SAM_OUT1[:-8])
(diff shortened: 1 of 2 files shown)


idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -331,97 +339,99 @@ def fetch_genome_work(s3genome, strict):
...
-            # something does not go as planned, because of the prefetching pattern of use.  Also, this function
-            # will fetch whichever of the "tar" or "tar.gz" versions of the archive actually exist in s3,
-            # even if the other version has been specified.  So the two functions are resilient in different ways.
             install_s3mi()
             tarfile = uncompressed(s3genome)
             try:
-                execute_command(
-                    "s3mi cat {tarfile} | tar xvf - -C {refdir}".format(
-                        tarfile=tarfile, refdir=REF_DIR))
+                print("Trying to download compressed genome...")
```
jameshowardwang: should these be using logger / write_to_log?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -331,97 +339,99 @@ def fetch_genome_work(s3genome, strict):
                            "hg38_pantro5_k16"):
         write_to_log("Oh hello interesting new genome {}".format(genome_name))
     genome_dir = os.path.join(REF_DIR, genome_name)
+
     if not os.path.exists(genome_dir):
         try:
-            #TODO Clean up and fold into fetch_reference
-            # Hmm. I understand the desire to create a single function for fetching any reference we might need,
-            # but the functionality of these two functions is actually quite different.  One of them handles
```
jameshowardwang: You seem to be deleting these comments without actually implementing what they are suggesting. Is this because we no longer think the recommendation in the TODO is valid?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -583,15 +605,20 @@ def unmapped_files_in(some_dir):
...
         if f is not None:
-            output_i = os.path.join(RESULT_DIR, star_outputs[i])
+            output_i = result_path(star_outputs[i])
             execute_command("mv %s %s;" % (f, output_i))
             uploader_start(output_i, SAMPLE_S3_OUTPUT_PATH + "/")
-    # cleanup
+    # Cleanup
     execute_command("cd %s; rm -rf *" % SCRATCH_DIR)
-    write_to_log("finished job")
+    print("Finished running STAR.")
```
jameshowardwang: why did you change this?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -613,120 +641,131 @@ def run_priceseqfilter(input_fqs, uploader_start):
...
-        uploader_start(
-            os.path.join(RESULT_DIR, LZW_OUT2), SAMPLE_S3_OUTPUT_PATH + "/")
-    write_to_log("finished job")
+        uploader_start(result_path(LZW_OUT2), dst)
+    print("Finished running LZW.")
 
 
 def run_bowtie2(input_fas, uploader_start):
-    # check if genome downloaded already
+    """Bowtie is an aligner used for filtering out host genomes. Two input
```
cdebourcy: Love all the clarifying comments regarding external tools. I would change the first sentence of this one to "Bowtie2 is an aligner we use for filtering out reads that map to the host genome"

idseq_pipeline/commands/host_filtering_functions.py
jshoe: I believe they actually do the same thing throughout the program after the logger is configured, but write_to_log is probably better / changed it

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -735,12 +774,19 @@ class SkipGsnap(Exception):
 
 
 def run_gsnap_filter(input_fas, uploader_start):
+    """GSNAP is an aligner used for filtering out host genomes. Two input
```
cdebourcy: Would change to "GSNAP is an aligner we use to filter out reads mapping to the host genome."
Would also add why we do this additional step after STAR/bowtie-based filtering: to increase our sensitivity for host filtering. We currently only do this for the human host -- we use the chimp reference genome as an extension to increase our sensitivity for human matches

idseq_pipeline/commands/host_filtering_functions.py
jshoe: I took this out b/c the "Hmm" part was Boris responding to the TODO (to not combine with fetch_reference), and it seems like they do do pretty different things so I don't think we should combine them. But I could leave it there?

idseq_pipeline/commands/host_filtering_functions.py
```diff
@@ -124,6 +124,10 @@ def fq2fa(input_fastq, output_fasta):
     execute_command(cmd)
 
 
+def result_path(basename):
```
boris-dimitrov: There is (or was?) a similar result_dir function in non_host_alignment_functions.py, if I recall correctly.

idseq_pipeline/commands/host_filtering_functions.py
boris-dimitrov: Yeah this was a discussion between YF and me in the comments.   I am not sure if we need it or not need it;  but if code cleanliness is an objective, we could record the discussion in an issue and refer to it here, shortening the comment.

idseq_pipeline/commands/host_filtering_functions.py
jshoe: Ah ok I changed result_dir to be called result_path but I think I'll leave the declarations separate since common.py doesn't get the RESULT_DIR (SAMPLE_DIR) declarations

boris-dimitrov: Sounds good.

PR #224, opened 2018-06-05
(Minor) Changes to accessionid2seq_functions mostly

jshoe:
- Test run for this PR: https://staging.idseq.net/samples/658
- Bottom part is just changing some print statements to write_to_log (although it's not used in accessionid2seq_functions)


idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -353,55 +370,59 @@ def get_sequence_by_accession_id_ntf(accession_id, nt_loc_dict, ntf):
...
-                break
-            except:
-                if attempt + 1 < NUM_RETRIES:
-                    time.sleep(1.0 * (4**attempt))
-                else:
-                    print("All retries failed for get_sequence_by_accession_id_s3.")
-                    raise
-            finally:
-                try:
+                if pipe_file != '':
```
cdebourcy: is the goal of this new check to get rid of try/except here?

idseq_pipeline/commands/common.py
```diff
@@ -640,7 +640,8 @@ def run_and_log_work(log_params, target_outputs, lazy_run, func_name,
     global OUTPUT_VERSIONS
     global run_and_log_mutex
 
-    write_to_log("========== %s ==========", log_params.get("title"))
+    msg = "========== %s ==========" % log_params.get("title")
```
tfrcarvalho: Maybe not to be addressed on this PR, but why are we not using logger directly?
There are multiple benefits: you can use handlers to output to multiple places, you do not have to process the string (i.e. '%') and you can be more flexible with the log levels.

idseq_pipeline/commands/accessionid2seq_functions.py
jshoe: Oh it was actually because I got a linter warning that pipe_file could be undeclared so I instantiated pipe_file = '' above the block and added this line. But pipe_file is the first line in the 'try' so I took this out again.

idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -18,36 +18,40 @@
...
-                    if species_id > 0 or species_id < INVALID_CALL_BASE_ID:  # it's a match
-                        ma = re.search(adv_search_string, read_id)
-                        if ma:
-                            read2seq[ma.group(4).rstrip()] = [
-                                sequence.rstrip(),
-                                ma.group(1),
-                                ma.group(2),
-                                ma.group(3)
-                            ]
+                continue
```
boris-dimitrov: may be just this once let's nest

idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -138,18 +149,19 @@ def generate_alignment_viz_json(nt_file, nt_loc_db, db_type, annotated_m8,
...
                 ref_end = int(metrics[-3])
                 if ref_start > ref_end:  # SWAP
                     (ref_start, ref_end) = (ref_end, ref_start)
                 ref_start -= 1
-                prev_start = (ref_start - REF_DISPLAY_RANGE
-                              ) if (ref_start - REF_DISPLAY_RANGE) > 0 else 0
+
+                prev_start = 0
+                if (ref_start - REF_DISPLAY_RANGE) > 0:
+                    prev_start = (ref_start - REF_DISPLAY_RANGE)
```
boris-dimitrov: Since 0 is not the default or a common case, I think the code is not more readable by putting it first.   But I also think the version before your change wasn't as readable as it should have been.  In my view the best style here for Python is
```
    prev_start = ref_start - REF_DISPLAY_RANGE
    if prev_start < 0:
        prev_start = 0
```

idseq_pipeline/commands/accessionid2seq_functions.py
```diff
@@ -353,55 +370,57 @@ def get_sequence_by_accession_id_ntf(accession_id, nt_loc_dict, ntf):
...
+            seq_name = subprocess.check_output(
+                cmd, executable='/bin/bash', shell=True).split(" ", 1)[1]
+            exitcode = get_range_proc.wait()
+            msg = "Error in get_sequence_by_accession_id_s3."
+            assert exitcode == 0, msg
+            seq_len = os.stat(accession_file).st_size
+            break
+        except:
+            if attempt + 1 < NUM_RETRIES:
+                time.sleep(1.0 * (4**attempt))
```
boris-dimitrov: can this benefit from the retry decorator?   i am not sure, just asking

idseq_pipeline/commands/non_host_alignment_functions.py
```diff
@@ -124,17 +124,15 @@ def result_dir(basename):
...
 
 
 def count_lines_in_paired_files(input_files):
     distinct_counts = list(set(map(count_lines, input_files)))
-    assert len(
-        distinct_counts
-    ) == 1, "Mismatched line counts in supposedly paired files: {}".format(
+    msg = "Mismatched line counts in supposedly paired files: {}".format(
         input_files)
+    assert len(distinct_counts) == 1, msg
```
boris-dimitrov: There is no need to compute the message if the assert condition passes, which is the common case.   In this refactor, however, the message is always computed.   I am curious if python includes the optimization I am assuming exists.

boris-dimitrov:
I found one thing to change, and for the other thing it's up to you. :)

idseq_pipeline/commands/accessionid2seq_functions.py
jshoe: hm yeah that might be good

PR #228, opened 2018-06-19
Put in Boris's version of count_reads

jshoe:
- Test run for this: https://staging.idseq.net/samples/834

### DIFF
--- a/idseq_pipeline/commands/common.py
+++ b/idseq_pipeline/commands/common.py
@@ -492,6 +492,7 @@ def count_reads(file_name, file_type, max_reads=None):
             if should_count(line):
                 count += step
             if max_reads is not None and count > max_reads:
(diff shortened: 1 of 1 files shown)


idseq_pipeline/commands/common.py
```diff
@@ -458,33 +458,39 @@ def percent_str(percent):
...
+    count = 0.0
+    step = COUNTING_LOGIC[file_type]['step']
+    should_count = COUNTING_LOGIC[file_type]['should_count']
+    file_opener = open
+    if file_name.endswith(".gz"):
+        file_opener = gzip.open
+    with file_opener(file_name) as f:
+        for line in f:
+            if should_count(line):
+                count += step
```
boris-dimitrov: Could you also include the max_reads code that I accidentally omitted?

PR #229, opened 2018-06-19
Fix fasta_paired hardcoding

jshoe:
- TEST RUN: https://staging.idseq.net/samples/836
- Initial file type gets set as like:
```
    initial_file_type_for_log = "fasta"
    if "fastq" in FILE_TYPE or "fq" in FILE_TYPE:
        initial_file_type_for_log = "fastq"
    if len(fastq_files) == 2:
        initial_file_type_for_log += "_paired"
```


jshoe:
^ going to change them all to either fasta or fasta_paired

